PR #1139, opened 2026-03-12
feat: agentic workflow platform with planner, loops, HTTP, code, triggers & visual editor

manzke:
## Summary

- **8 new node types**: planner (LLM-based task planning with sub-workflow execution), verifier (quality gate with retry loops), loop (for/forEach/while), parallel + join (concurrent branches), HTTP request (with SSRF protection), code execution (VM-sandboxed JavaScript)
- **Trigger system**: cron scheduling via `croner` and webhook triggers with HMAC-SHA256 verification
- **Visual workflow editor**: ReactFlow-based drag-and-drop builder with dagre auto-layout, node palette, config panel, and version control (publish/activate)
- **Context summarization**: Automatic LLM-based summarization when workflow state exceeds token threshold
- **Sub-workflow execution**: Planner nodes dynamically create and execute child workflows with depth limits and abort signal propagation

### Architecture

Uses **Option C** (dynamic sub-workflows): Planner nodes materialize child workflow definitions at runtime, executed via the existing `WorkflowEngine.start()`. Parent DAG immutability is preserved — all dynamic behavior happens inside isolated child executions.

### Security

- VM sandbox: null-prototype context + JSON round-trip + strict mode + Function disabled
- SSRF protection: blocks private IPs (10.x, 172.16-31.x, 192.168.x, 127.x, 169.254.x, IPv6 ULA/link-local)
- Webhook auth: HMAC-SHA256 with `crypto.timingSafeEqual`
- Sub-workflow depth limits (default 3)
- Code execution timeout enforcement

### Files changed

38 files, +4,242 lines across server (executors, triggers, routes, engine) and client (editor, SSE events, UI updates)

## Test plan

- [ ] Verify server starts: `timeout 8s node server/server.js`
- [ ] Verify client builds: `cd client && npm run build`
- [ ] Verify lint clean: `npm run lint:fix` (0 errors)
- [ ] Test planner node creates sub-workflow with task plan
- [ ] Test loop node iterates correctly (for/forEach/while modes)
- [ ] Test HTTP node with SSRF protection rejects private IPs
- [ ] Test code node sandbox blocks `process`, `require`, `Function`
- [ ] Test webhook trigger with HMAC signature verification
- [ ] Test visual editor: create workflow, add nodes, connect, save
- [ ] Test version control: publish and activate workflow versions


manzke:
## Review summary

Did a careful pass across simplicity, code, architecture, security, and product. The architecture (Option C: planner materializes child workflows, parent DAG immutable) is sound, the executor abstraction is clean, and the test plan in the description is sensible. But there are several **runtime-breaking bugs and exploitable security gaps** that block merging as-is.

I've pushed fixes for all of them on top of this branch: **#1476** (targets `feature/agentic-workflows-v2`). Highlights:

### Critical (broken at runtime)

1. **Trigger endpoints always return 400.** `validateIdForPath()` returns a boolean but `server/routes/workflow/triggerRoutes.js:71-114` treats it as `{valid, sanitized}`. Both `POST /api/workflows/:id/trigger` and the public webhook endpoint are unusable in their current state.
2. **Removed triggers keep firing.** `TriggerManager.registerWorkflowTriggers()` returns early when the new triggers list is empty, so it never unregisters the previously registered triggers. Updating a workflow to remove all triggers leaves them running until process restart.
3. **Webhook id collisions.** Webhook map is keyed by `triggerConfig.id` alone. Two workflows that each define `{ id: 'gh', type: 'webhook' }` will alias — the later registration overwrites the earlier, and the workflow check returns 404 for the wrong workflow.
4. **Client subscribes to events the server filters out.** `useWorkflowExecution.js` listens for `workflow.plan.created` and `workflow.subworkflow.{start,complete}`, but the server's SSE allowlist in `workflowRoutes.js:1297-1310` doesn't include them, so `planCreated` and `subworkflows` UI state never populates.
5. **"New Workflow" button links to the wrong page.** `/admin/workflows/new` hits the legacy JSON editor (`AdminWorkflowEditPage`), not the new visual one. Should be `/admin/workflows/new/edit`.

### Security

6. **Webhook HMAC uses re-stringified body.** `WebhookTrigger.verifySignature()` computes HMAC over `JSON.stringify(payload)`. Real webhook providers (GitHub, Stripe, etc.) sign the **raw bytes** they sent. Parsing + re-stringifying changes byte order, so valid third-party signatures are rejected. Fix is to capture raw bytes with `express.raw()` on the route and verify against them.
7. **Secret-less webhook accepts any payload.** `verifySignature` returns `true` when no secret is configured. Combined with #8 below, anyone who can reach the server can execute any workflow with a webhook trigger.
8. **Webhook executions run as admin.** `TriggerManager.fireTrigger()` hardcodes `user: { id: 'system', name: 'System Trigger', groups: ['admin'] }`. Even scheduled triggers — and especially webhook-fired ones — should not have admin privileges by default. Fix: synthetic user with no group memberships; manual-trigger endpoint passes through `req.user`.
9. **Path traversal in version control endpoints.** `req.params.version` is validated with `/^[\w.-]+$/`, which **allows `..`** and other path-traversal sequences. The match is then used in `f.startsWith(versionParam)` against `fs.readdir` output — not directly traversal, but the regex is wrong and CodeQL is right to flag it. Tightened to alphanumeric-anchored ends + explicit `..` ban, and added `resolveAndValidatePath()` defense-in-depth for the history dir and snapshot path.
10. **HttpNodeExecutor SSRF is bypassable.** The hostname regex check misses:
    - **Domain names that resolve to private IPs** (e.g. `internal.attacker.com → 127.0.0.1`)
    - **IPv4-mapped IPv6** (`::ffff:127.0.0.1`)
    - **Bracketed IPv6** (the URL parser leaves `[::1]` with brackets, so `hostname === '::1'` is false)
    - **Multicast/reserved IPv4** (`224.0.0.0/4`, `240.0.0.0/4`)
    - **Redirects** — fetch follows them by default, so a public server can 302 to `169.254.169.254` after the SSRF check passes (AWS IMDS data exfiltration).
    - **Non-http(s) schemes** like `file:`, `gopher:`, `ftp:`.
    Fix resolves A+AAAA records and checks every resolved IP, uses `redirect: 'manual'`, and rejects non-http(s) schemes.

### Code sandbox limitations (documented, not fixed)

11. `CodeNodeExecutor` and `LoopNodeExecutor.evaluateCondition` use `vm.runInNewContext` with a null-prototype context, JSON round-trip, and `Function = undefined`. This is reasonable hardening but is **not a security boundary**: Promise microtasks bypass the timeout, exposed primitives can leak access to internal Function constructors in some Node versions, and resource exhaustion is still possible. As long as **only admins can author workflows** with code/loop nodes, this is acceptable. If non-admins ever become authors, you'll need isolated-vm or worker-thread sandboxing. I'd add a code comment to that effect.

### Style / minor

12. **i18n.** `prompt('Enter workflow ID:')`, `alert('Save failed: …')`, and several hardcoded strings in the visual editor are not localized. Added the `workflows.*` keys (en + de) and replaced the `alert()` calls with an in-app toast.
13. **Atomic writes.** `publish` and `activate/:version` use `fs.writeFile(... JSON.stringify(...) ...)` directly; the rest of the codebase uses `atomicWriteJSON`. Switched both to atomic.

### Out of scope for this review

- The PR is **far behind `main`** (746 files diverged — Bedrock, Nextcloud, compare mode, Office365 fixes, etc.). It'll need a rebase before merging; I haven't attempted one because the conflicts cross too many subsystems.
- No integration tests for the new node executors. The existing `tests/integration/workflows/` patterns would extend naturally; worth adding before this becomes load-bearing.

### Tests added in #1476

- `WebhookTrigger.verifyRawSignature`: valid prefixed/unprefixed sigs, length-mismatch (no throw), missing signature, secret-less trigger never accepting, re-stringified body producing a different signature than raw body.
- Hand-verified SSRF table (20/20) for IPv4/IPv6/private/public/IPv4-mapped.
- Hand-verified version regex (20/20) against path-traversal inputs.
- `TriggerManager` composite-key isolation + always-unregister-on-update verified.

Happy to iterate on any of the above — particularly the SSRF hardening, where I went with DNS pre-resolution + manual redirects, but a custom HTTP agent with a custom dispatcher would be stronger if you want belt-and-suspenders.


---

PR #1228, opened 2026-04-01
fix(a11y): WCAG 2.1 Level AA keyboard navigation improvements

manzke:
Implements comprehensive WCAG 2.1 Level AA accessibility improvements for keyboard navigation across chat interface components.

### ChatMessage Action Menu
- Added `onFocus` and `onBlur` handlers to message container to show action menu when keyboard focused
- Action menu (feedback, download, etc.) now visible when navigating via Tab key

### ModelSelector Keyboard Navigation
- Implemented roving tabindex pattern with `useKeyboardNavigation` hook
- ESC key closes the dropdown
- Arrow up/down keys navigate through model options
- Enter/Space keys select the focused model and open/close dropdown
- Fixed event capture phase to prevent container scrolling during arrow key navigation
- Added proper ARIA roles (`role="menu"`, `role="menuitem"`, `aria-haspopup`, `aria-expanded`)
- Added visible focus indicators to all model options

### StarRating Keyboard Accessibility
- Converted star elements from `<div>` to `<button>` for proper keyboard support
- Added keyboard handlers for Enter/Space to select ratings
- Added `tabIndex`, focus styling (`focus:ring-2 focus:ring-indigo-500`), and `aria-label` attributes
- Stars now fully navigable and selectable via keyboard

### Keyboard Navigation Hook Enhancement
- Modified `useKeyboardNavigation` hook to use event capture phase
- Prevents browser's default scroll behavior from interfering with keyboard navigation
- Ensures arrow key handlers fire before scrolling occurs in scrollable containers

## Technical Implementation

- **Roving Tabindex Pattern**: Active item has `tabIndex={0}`, inactive items have `tabIndex={-1}`
- **Event Capture Phase**: `addEventListener('keydown', handler, true)` intercepts events before bubble phase
- **Focus Management**: Automatic focus and visual indicators on all interactive elements
- **Screen Reader Support**: Proper ARIA attributes for announcements and roles

## Testing

- ✅ Tab navigation shows all hidden menus and actions
- ✅ ModelSelector opens with Enter/Space, navigates with arrows, closes with ESC
- ✅ Arrow keys navigate models without scrolling the container
- ✅ StarRating fully keyboard accessible with Enter/Space
- ✅ All interactive elements have visible focus indicators
- ✅ ESLint 10 compatibility maintained

All changes follow WCAG 2.1 Level AA guidelines for keyboard accessibility.


manzke (commented):
@claude[agent] the hidden menus under the chat messages are not shown, when selected via tab / accessibility. you can iterate through them, but they are not shown, so a user does not that they are selected. the model selector also works strange. first it can't be closed with ESC. second, when selected via tab and enter to open, you can't navigate them with arrow up and down. you have to use tab instead.

PR #1396, opened 2026-05-08
fix(oidc): read groups claim from ID token

manzke:
## Summary

Some OIDC providers — most notably **Microsoft Entra** — emit the `groups` claim only in the ID token, not in the userinfo endpoint response. Today the verify callback in `server/middleware/oidcAuth.js` declares 4 arguments, so [passport-oauth2 dispatches](https://github.com/jaredhanson/passport-oauth2/blob/master/lib/strategy.js#L194-L205) without the `params` argument and `id_token` is silently dropped on the floor. Result: even with `groupsAttribute: "groups"` configured, no group ever reaches `user.externalGroups`, so the `groups.json` mapping never fires and every authenticated user effectively shares the same role.

This change makes the OIDC middleware also read groups from the ID token claims.

## What changes

- `server/middleware/oidcAuth.js`
  - Verify callback signature: `(accessToken, _refreshToken, _profile, done)` → `(accessToken, _refreshToken, params, _profile, done)`.
  - Decode `params.id_token` via the new `decodeIdTokenClaims()` helper.
  - When `userInfo[provider.groupsAttribute]` is missing but the ID token has it, merge the array into `userInfo` so the existing extraction at the bottom of the callback stays unchanged.
  - Log a structured `id_token_claims_decoded` debug event and a `WARN` when Entra's `_claim_names.groups` / `hasgroups` overage indicator is present (so we notice it instead of silently truncating).
- `server/utils/oidcIdToken.js` (new): tiny dependency-free `decodeIdTokenClaims()` helper. Signature verification is intentionally skipped — the token came from the IdP token endpoint over TLS to passport-oauth2 in this same process and is used here only as a side-channel alongside userinfo.
- `tests/unit/server/oidcAuth.test.js` (new): unit tests for the decoder, including the Entra overage shape and malformed-input cases.

No config schema or migration needed — all the levers (`scope`, `groupsAttribute`, `defaultGroups`) already exist in `oidcProviderSchema`. For Entra specifically, the App Registration still needs **Token configuration → Add groups claim** to actually emit groups in the ID token.

## Why no signature verification on the ID token

Doing it properly requires fetching the IdP's JWKS, picking a JWS verifier, and handling key rotation — all out of scope for this fix. The token reaches us directly from `passport-oauth2` after a TLS exchange with the token endpoint, so for the purpose of populating `externalGroups` (which is then run through the existing `groups.json` allow-list anyway), unverified decoding is acceptable. A follow-up could swap in `openid-client` or `passport-openidconnect` for full verification.

## Out of scope (deliberate)

- **Microsoft Graph fallback for the overage case.** When a user is in >200 security groups, Entra emits `_claim_names` instead of the full `groups` array. We log a warning so it's visible in pod logs, but we don't yet call `/me/transitiveMemberOf`. Reference scripts already exist outside the repo (`entra-test.js`) and a follow-up PR can wire this in if any tenant actually hits the limit.
- **Tightening default group permissions.** `users` and `authenticated` currently both grant `apps: ["*"]`, so even after groups start flowing, every user keeps seeing every app. That's a configuration change, not a code change, and belongs in `groups.json` per deployment.

## Test plan

- [x] `npm run test:unit -- tests/unit/server/oidcAuth.test.js` — 6/6 passing locally
- [x] `npx eslint server/middleware/oidcAuth.js server/utils/oidcIdToken.js tests/unit/server/oidcAuth.test.js` — clean
- [x] `npx prettier --check` on the same files — clean
- [x] `node server/server.js` boots without errors
- [ ] **Manual on a real Entra tenant**: add the groups claim in App Registration → Token configuration, log in, watch pod logs for `groups_merged_from_id_token` with a non-zero count, and confirm a non-admin user that's added to an `admins`-mapped Entra group via GUID flips to admin in the UI

### DIFF
--- a/server/middleware/oidcAuth.js
+++ b/server/middleware/oidcAuth.js
@@ -9,6 +9,7 @@ import authDebugService from '../utils/authDebugService.js';
 import logger from '../utils/logger.js';
 import { buildServerPath } from '../utils/basePath.js';
 import { getAuthCookieOptions } from '../utils/cookieSettings.js';
+import { decodeIdTokenClaims } from '../utils/oidcIdToken.js';
 
 // Store configured providers
 const configuredProviders = new Map();
@@ -69,7 +70,7 @@ export function configureOidcProviders() {
           customHeaders: {},
           skipUserProfile: false // We'll fetch user info manually
         },
-        async (accessToken, _refreshToken, _profile, done) => {
+        async (accessToken, _refreshToken, params, _profile, done) => {
           const sessionId = authDebugService.generateSessionId();
 
           try {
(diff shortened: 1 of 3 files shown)


manzke:
@copilot apply changes based on the comments in [this thread](https://github.com/intrafind/ihub-apps/pull/1396#pullrequestreview-4251367465)

PR #1523, opened 2026-05-22
feat(i18n): allow parent frames to set language via postMessage

manzke:
## Summary

When iHub is embedded as an iframe, the host page had no good way to drive the UI language at runtime:

- `navigator.language` only wins on the **first** visit (no localStorage yet). Once `localStorage.i18nextLng` is set, it sticks across reloads — even if the parent page is now in a different language.
- The existing `?language=de` URL parameter works at load time but cannot react to language switches in the parent without reloading the iframe.

This PR adds a `postMessage` channel so the parent window can switch iHub's language at any time, persisting the new value to `localStorage` (so it survives reloads in the same iframe).

## How it works

The i18n service registers a `message` listener on `window` during init and watches for:

```javascript
{ type: 'ihub:setLanguage', language: 'de' }
```

Flow:
1. Payload validated against a BCP 47 basic pattern (`en`, `de`, `en-US`, `pt-BR`).
2. Invalid messages are dropped with a console warning. Any message of a different `type` is ignored, so the listener is safe to leave running on pages that exchange unrelated postMessage traffic.
3. Valid changes go through the existing `changeLanguage()` flow, which updates i18next *and* writes `localStorage.i18nextLng` via the detector's `caches: ['localStorage']` setting.
4. The iframe replies to `event.source` with `{ type: 'ihub:languageChanged', language }` once the change is applied.

Detection priority is unchanged for organic visits:
1. `?language=de` URL parameter
2. `localStorage.i18nextLng` (the user's last explicit choice)
3. `navigator.language` (first visit only)
4. server `defaultLanguage` fallback

## Parent-side usage

```javascript
const iframe = document.getElementById('ihub-iframe');
iframe.contentWindow.postMessage(
  { type: 'ihub:setLanguage', language: 'de' },
  'https://ihub.example.com'
);

window.addEventListener('message', event => {
  if (event.data?.type === 'ihub:languageChanged') {
    console.log('iHub now in', event.data.language);
  }
});
```

## Files changed

- `client/src/services/i18nService.js` — added `setupPostMessageListener()` / `handlePostMessage()` and wired the listener into `initializeSync()`.
- `docs/localization.md` — documented the client-side detection priority and the new postMessage API.

## Test plan

- [ ] Fresh browser profile, German UI language: load iHub in an iframe with no params → UI renders in German (existing behavior, verified unchanged).
- [ ] After a language switch via the in-app selector, reload the iframe → UI keeps the user's choice (existing behavior, verified unchanged).
- [ ] From parent, `postMessage({ type: 'ihub:setLanguage', language: 'de' })` while iframe is showing English → UI flips to German without reload.
- [ ] Verify `localStorage.i18nextLng` is updated to `de` after the postMessage.
- [ ] Reload the iframe with no params → UI stays in German (persisted).
- [ ] Send invalid payloads (`{ language: '../etc' }`, `{ type: 'other' }`, no language) → ignored, console warns where appropriate, no exceptions.
- [ ] Parent listens for `ihub:languageChanged` ack → receives it after the change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QgYSPfYit6vGbRvQeduGCg)_

### DIFF
--- a/client/src/services/i18nService.js
+++ b/client/src/services/i18nService.js
@@ -60,12 +60,63 @@ class I18nService {
 
       this.isInitialized = true;
 
+      // Allow parent frames to drive the language via postMessage
+      this.setupPostMessageListener();
+
       // Load full setup asynchronously
       this.initializeAsync();
     } catch (error) {
       console.error('Failed to initialize i18n service synchronously:', error);
       this.isInitialized = true;
     }
+  }
+
+  setupPostMessageListener() {
+    if (typeof window === 'undefined' || typeof window.addEventListener !== 'function') {
+      return;
+    }
+
+    window.addEventListener('message', event => this.handlePostMessage(event));
+  }
+
+  handlePostMessage(event) {
+    const data = event?.data;
+    if (!data || typeof data !== 'object') return;
+    if (data.type !== 'ihub:setLanguage') return;
+
+    const requestedLanguage = data.language;
+    if (typeof requestedLanguage !== 'string' || !requestedLanguage.trim()) {
+      console.warn('[i18n] postMessage ihub:setLanguage ignored: missing language');
+      return;
+    }
+
+    // Accept BCP 47 basic forms (e.g. "en", "de", "en-US", "pt-BR")
+    if (!/^[A-Za-z]{2,3}(-[A-Za-z]{2,4})?$/.test(requestedLanguage.trim())) {
+      console.warn(
+        `[i18n] postMessage ihub:setLanguage ignored: invalid language code "${requestedLanguage}"`
+      );
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


manzke (commented):
@claude[agent] make sure to use and listen for "Channel and message conventions
Channel name: 'intrafind' — product-neutral so iFinder, iAssistant, and any other IntraFind product can publish/subscribe on the same channel. Confirmed there is no pre-existing BroadcastChannel in iFinder; iAssistant will need to adopt the same name on its receiving side.
Message shape: { type: 'language-changed', language: '<locale>' }. The type field is unprefixed because the channel name already provides the namespace. Future cross-product events use the same channel and discriminate via type."

PR #1876, opened 2026-07-07
fix(outlook): stop silently dropping unsupported email attachments

### LINKED ISSUE #1451: [Outlook] Bug: PDF / Word / .eml attachments fail silently or throw
## Summary

**Bug.** Several users report that opening the add-in on emails that contain PDF, Word, or message (`.eml`/`.msg`) attachments produces an error. The attachment-handling path currently swallows errors silently and falls over on certain MIME types instead of degrading gracefully.

## Current behavior

`client/src/features/office/utilities/outlookMailContext.js` (lines 109–129):

```js
for (const d of descriptors) {
  if (!d.id) {
    attachments.push({ ...d, error: 'Missing attachment id' });
    continue;
  }
  try {
    const raw = await getAttachmentContentAsync(d.id);
    attachments.push({ ...d, content: { format: raw.format, content: raw.content } });
  } catch (e) {
    attachments.push({ ...d, error: e?.message ?? String(e) });
  }
}
```

`getAttachmentContentAsync` (lines 33–48) rejects with "getAttachmentContentAsync is not available (requires Mailbox 1.8+)" on older hosts.

`client/src/features/office/utilities/buildChatApiMessages.js` (lines 43–69):

```js
files.map(async a => {
  try {
    const file = base64ToFile(a.content.content, a.name, a.contentType);
    const { content, pageImages } = await processDocumentFile(file);
    return { ... };
  } catch {
    return null;          // 👈 silently dropped, no log, no user feedback
  }
});
```

Problems:
- `processDocumentFile` is not guarded by MIME type — it gets called on every non-image attachment, including types it can't handle (e.g. `.msg`, `.eml`, encrypted PDFs, `.zip`).
- Failures inside `processDocumentFile` are silently caught and converted to `null`, so the user gets neither the attachment nor any indication why.
- `format: 'eml'` items returned by `getAttachmentContentAsync` (for attached emails) flow through `base64ToFile` even though their `content` is a base64-encoded MIME message, not a binary doc. This is a likely root cause of the email-attachment errors.
- No correlation id / file name in error logs, making the bug hard to reproduce from logs.

Related: `concepts/2026-04-22 Outlook File Upload Fix.md`.

## Repro

1. Open an email in Outlook desktop or OWA with a PDF, a Word doc, **and** a forwarded `.eml` attachment.
2. Open the add-in.
3. Observe: an error notification appears, or the chat starts without the attachments and no explanation is given.

## Proposed change

1. **Pre-filter by MIME / format before `processDocumentFile`:**
   - Whitelist: `application/pdf`, common Office MIME types (`application/vnd.openxmlformats-…`, `application/msword`, etc.), plain text.
   - Skip (with a logged reason): `format: 'eml'` attachments and `attachmentType === 'item'` — these need a dedicated handler.
2. **Surface failures to the user**: change the silent `catch { return null }` to return a `{ fileName, error }` entry. Render unsupported attachments in the upcoming review banner (see "Review attachments" issue) with a clear "Unsupported / failed" status, instead of dropping them.
3. **Log structured errors** including attachment name, content type, format, and the underlying error message. Send to the existing client error reporter.
4. **Handle attached emails (`format: 'eml'`)** explicitly: either decode them server-side, or extract their text/subject/from and inline as quoted context. At minimum, fail gracefully with a user-visible message ("Email attachments are not yet supported").
5. **Mailbox API ≥ 1.8 check**: when `getAttachmentContentAsync` is not available, show a single banner rather than a per-attachment error per call.

## Acceptance criteria

- [ ] Opening an email with PDF + Word + .eml attachments no longer throws or blocks the chat.
- [ ] Each attachment renders with one of: ✅ "attached", ⚠️ "unsupported (eml)", ❌ "failed: <reason>".
- [ ] Errors include the attachment name and content type in the client log.
- [ ] Image attachments still flow through `buildImageDataFromMailAttachments` unchanged.
- [ ] A regression test covers at least: PDF, .docx, .eml, and an oversized attachment.

## Files

- `client/src/features/office/utilities/outlookMailContext.js`
- `client/src/features/office/utilities/buildChatApiMessages.js`
- `client/src/features/upload/utils/fileProcessing.js` (where `processDocumentFile` lives)
- `concepts/2026-04-22 Outlook File Upload Fix.md`

## Source

User feedback (multiple), 2026-05-13.

manzke:
## Summary

Fixes #1451.

Attachments to an Outlook email that Office.js returns as attached-email
(`format: 'eml'`), meeting-invite (`format: 'icalendar'`), or cloud-share-link
(`format: 'url'`) content fetch successfully — so they have no `.error` — but
were previously treated as unusable and silently dropped by
`buildFileDataFromMailAttachments` via `hasBase64Content`, with no log line
and no indication to the user that anything was missing, even though the
review banner showed them as "attached".

Rather than just detecting and flagging these as "unsupported" (this PR's
first iteration), `.eml` and `.icalendar` are plain-text formats — so they're
now actually parsed into real content instead.

## Changes

- **`client/src/features/office/utilities/emailAttachmentParsers.js`** (new):
  dependency-free parsers —
  - `parseEmlAttachment`: decodes an attached/forwarded email into a
    "Subject/From/To/Cc/Date + body" text block (handles multipart/alternative,
    quoted-printable and base64 `Content-Transfer-Encoding`, RFC 2047 encoded
    headers, and falls back to a regex-based HTML→text conversion for
    HTML-only bodies).
  - `parseIcsAttachment`: decodes a meeting invite into a short summary
    (subject, start/end time, location, organizer).
  - Pure (`atob`/`TextDecoder`/regex only, no DOM) so it's unit-testable under
    plain Node.
- **`buildChatApiMessages.js`**: `buildFileDataFromMailAttachments` now
  dispatches on the attachment's actual content format — `eml`/`icalendar` go
  through the new parsers, `url` (OneDrive/SharePoint share links, which have
  no fetchable file bytes) sends the link itself as a text reference instead
  of being dropped, and everything else goes through the existing
  binary-document pipeline (PDF/DOCX/XLSX/PPTX/MSG/...) unchanged.
- **`client/src/features/office/utilities/attachmentFormat.js`** (new in this
  PR): `sanitizeContentType`/`hasBase64Content` extracted to a dependency-free
  leaf module (previously inline in `buildChatApiMessages.js`, which pulls in
  the full document pipeline) so they're unit-testable under plain Node.
- **`OfficeMailContextBanner.jsx`**: since every fetched, non-error attachment
  now produces real content, the banner's original attached/failed/pending
  states are enough again — no new "unsupported" status needed. Two small
  banner improvements remain from this PR: attachments over 20 MB are skipped
  up front (in `outlookMailContext.js`) instead of risking a hung task pane,
  and old Outlook hosts (pre-Mailbox 1.8, which can't fetch attachment content
  at all) get one explanatory banner line instead of the same error repeated
  per attachment.
- Regression tests (plain-Node, matching the existing `tokenStats.test.js`
  convention — no test runner is configured for the client):
  `emailAttachmentParsers.test.js` (eml/ics parsing, including malformed
  input) and `attachmentFormat.test.js` (content-format classification).

## Out of scope

- Per-attachment status still can't reflect a `processDocumentFile` failure
  that only happens at send time (e.g. an encrypted PDF) — the banner reflects
  what's known synchronously when the email is read. The failure is still
  logged and the attachment is dropped rather than sent as garbage.
- Cloud-hosted attachments (OneDrive/SharePoint `url` format) still can't have
  their actual file content retrieved client-side — that would need a Graph
  API integration with its own auth flow. They now send their share link as a
  reference instead of nothing.

## Test plan

- [x] `node client/src/features/office/utilities/emailAttachmentParsers.test.js` — all pass
- [x] `node client/src/features/office/utilities/attachmentFormat.test.js` — all pass
- [x] `node --check` on all modified/added plain-JS files
- [ ] Manual verification in Outlook (desktop/OWA) with an email containing a PDF, a Word doc, a forwarded `.eml`, and a meeting invite — not run in this environment (no Outlook host available); please verify before merging.

### DIFF
--- a/client/src/features/office/components/chat/OfficeMailContextBanner.jsx
+++ b/client/src/features/office/components/chat/OfficeMailContextBanner.jsx
@@ -1,7 +1,6 @@
 import { useMemo } from 'react';
 import Icon from '../../../../shared/components/Icon';
 import { formatFileSize } from '../../../upload/utils/cloudFileProcessing';
-import { isUnsupportedAttachmentFormat } from '../../utilities/attachmentFormat';
 import { MAILBOX_ATTACHMENT_API_UNAVAILABLE_MESSAGE } from '../../utilities/outlookMailContext';
 
 const IMAGE_EXT = /\.(png|jpe?g|gif|webp|bmp|svg)$/i;
@@ -16,9 +15,6 @@ function isImageAttachment(att) {
 
 function getAttachmentStatus(att) {
   if (att?.error) return { kind: 'failed', label: att.error };
-  if (isUnsupportedAttachmentFormat(att)) {
-    return { kind: 'unsupported', label: 'This attachment type is not supported yet' };
-  }
   if (att?.content) return { kind: 'attached', label: 'Will be sent' };
   return { kind: 'pending', label: 'Loading...' };
 }
@@ -211,14 +207,6 @@ function OfficeMailContextBanner({
                         <span aria-hidden>•</span>
                         <span className="text-rose-600" title={status.label}>
                           Failed
-                        </span>
-                      </>
-                    )}
-                    {status.kind === 'unsupported' && (
-                      <>
-                        <span aria-hidden>•</span>
-                        <span className="text-amber-600" title={status.label}>
-                          Unsupported
                         </span>
                       </>
                     )}
--- a/client/src/features/office/components/chat/OfficeMailContextBanner.jsx
+++ b/client/src/features/office/components/chat/OfficeMailContextBanner.jsx
@@ -1,6 +1,8 @@
 import { useMemo } from 'react';
 import Icon from '../../../../shared/components/Icon';
 import { formatFileSize } from '../../../upload/utils/cloudFileProcessing';
+import { isUnsupportedAttachmentFormat } from '../../utilities/attachmentFormat';
+import { MAILBOX_ATTACHMENT_API_UNAVAILABLE_MESSAGE } from '../../utilities/outlookMailContext';
 
 const IMAGE_EXT = /\.(png|jpe?g|gif|webp|bmp|svg)$/i;
 
@@ -14,6 +16,9 @@ function isImageAttachment(att) {
 
 function getAttachmentStatus(att) {
   if (att?.error) return { kind: 'failed', label: att.error };
+  if (isUnsupportedAttachmentFormat(att)) {
+    return { kind: 'unsupported', label: 'This attachment type is not supported yet' };
+  }
   if (att?.content) return { kind: 'attached', label: 'Will be sent' };
   return { kind: 'pending', label: 'Loading...' };
 }
@@ -64,6 +69,12 @@ function OfficeMailContextBanner({
 
   const hasBody = Boolean(ctx?.bodyText && ctx.bodyText.trim().length > 0);
   const hasAttachments = attachments.length > 0;
+  // Older Outlook hosts (pre-Mailbox 1.8) fail every attachment fetch with
+  // the same "API not available" error — show one explanation instead of
+  // repeating it on every row.
+  const attachmentApiUnavailable =
+    hasAttachments &&
+    attachments.every(a => a?.error === MAILBOX_ATTACHMENT_API_UNAVAILABLE_MESSAGE);
 
   if (loading) {
     return (
@@ -167,6 +178,14 @@ function OfficeMailContextBanner({
       {/* Attachments list */}
       {hasAttachments && remainingAttachments.length > 0 && (
         <div className="divide-y divide-slate-100">
+          {attachmentApiUnavailable && (
+            <div className="flex items-start gap-2 px-3 py-1.5 bg-amber-50 text-[11px] text-amber-700">
+              <Icon name="information-circle" size="sm" />
+              <span>
+                Attachments can&apos;t be read on this version of Outlook (requires Mailbox 1.8+).
+              </span>
+            </div>
+          )}
           {remainingAttachments.map(att => {
             const status = getAttachmentStatus(att);
             const isImage = isImageAttachment(att);
@@ -187,11 +206,19 @@ function OfficeMailContextBanner({
                   </div>
                   <div className="text-[11px] text-slate-500 flex items-center gap-1.5">
                     <span>{formatFileSize(Number(att.size) || 0)}</span>
-                    {status.kind === 'failed' && (
+                    {status.kind === 'failed' && !attachmentApiUnavailable && (
                       <>
                         <span aria-hidden>•</span>
                         <span className="text-rose-600" title={status.label}>
                           Failed
+                        </span>
+                      </>
+                    )}
+                    {status.kind === 'unsupported' && (
+                      <>
+                        <span aria-hidden>•</span>
+                        <span className="text-amber-600" title={status.label}>
+                          Unsupported
                         </span>
                       </>
                     )}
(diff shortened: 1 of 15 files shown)


manzke:
The `test-quick` failure (`tests/unit/server/locale-override.test.js`, `SyntaxError: Cannot use 'import.meta' outside a module`) is a pre-existing CI/Jest config issue, not caused by this PR:

- This PR doesn't touch that test file or any Jest/Babel config.
- The `Unit Tests` workflow (`.github/workflows/unit-tests.yml`) was added yesterday and is currently failing with the same error on essentially every other open branch I checked (dependabot, several `copilot/*` branches, other `claude/*` branches) — it looks like the shared Jest config doesn't yet handle `import.meta.url` in that one server test file.
- All 206 other tests in the suite pass; only this one suite fails to parse.

Not attempting a fix here since it's a repo-wide Jest config gap unrelated to the Outlook attachment fix in this PR — happy to pick it up separately if useful.

---

manzke:
Pushed a fix for the 3 CodeQL alerts on `stripHtmlToText` in `emailAttachmentParsers.js`:

- **Incomplete multi-character sanitization** (tag stripping): a single non-looping regex pass could leave a live `<script>` tag behind on crafted input like `<scr<script>ipt>...</scr</script>ipt>` (removing the inner match reassembles the outer one). Both the script/style block removal and the generic tag strip now loop to a fixed point.
- **Double escaping/unescaping**: sequential independent `&amp;` → `&lt;` → `<script>` etc. replacements could double-decode a safely double-escaped string. Entity decoding now runs in a single regex pass so each match resolves exactly once.

Added a regression test (`emailAttachmentParsers.test.js`) exercising both attack shapes plus the existing script/br/entity cases — confirms neither reconstructed tag nor double-unescape survives.

Also fixed the real `test-quick` failure: I'd missed that this exact module already has a Jest suite at `tests/unit/client/office-build-chat-api-messages.test.jsx` (the plain-node scripts under `client/src` are a separate, non-CI-enforced convention). Updated it for the new eml/icalendar/url dispatch behavior and added cases for the new code paths — 27/27 passing locally, plus the full `tests/unit` suite (209/209, excluding the pre-existing unrelated `locale-override.test.js` failure noted earlier).

---

manzke:
The latest `test-quick` run now fails 3 suites instead of 1 (`locale-override.test.js`, plus new failures in `pptx-file-extraction.test.jsx` and `msg-file-extraction.test.jsx`). This is **not caused by this PR** — it's a regression on `main` from #1862 (merged a few minutes ago, 09:35 UTC), which changed `client/src/features/upload/utils/fileProcessing.js` to:

```js
import pdfjsWorkerUrl from 'pdfjs-dist/build/pdf.worker.min.mjs?url';
```

The `?url` suffix is a Vite-only import convention — Jest's module resolver doesn't understand it and tries to open a literal file named `...pdf.worker.min.mjs?url`, which doesn't exist:

```
ENOENT: no such file or directory, open '.../client/node_modules/pdfjs-dist/build/pdf.worker.min.mjs?url'
```

Any test that transitively imports `fileProcessing.js` without mocking it now fails to load under Jest. I confirmed this reproduces locally against `main` directly (unrelated to my branch's diff), and it's already showing up as new `test-quick` failures on several other open PRs in this same window (e.g. `claude/beautiful-wozniak-qikpdj`, `claude/beautiful-wozniak-ooez61`).

Not fixing it here since it's out of scope for the Outlook attachment fix — flagging in case it needs a follow-up fix on `main` (e.g. a Jest `moduleNameMapper` entry for the `?url` suffix, or reverting to the `new URL(..., import.meta.url)` pattern from #1862's own description). The suite this PR actually touches (`office-build-chat-api-messages.test.jsx`, which mocks `fileProcessing.js`) is unaffected — 27/27 passing.

---

manzke:
@copilot please fix the failing tests even if they are a regression from main

PR #1878, opened 2026-07-07
Add Allowed Models selector to App Editor

### LINKED ISSUE #635: Check limit models for apps
It should be possible to configure / limit which models can be used for a specific app. Right now we can only configure which one is the preferred.

manzke:
## Summary

Closes #635 — "It should be possible to configure / limit which models can be used for a specific app. Right now we can only configure which one is the preferred."

Investigation showed `app.allowedModels` was already fully wired end-to-end except for one gap:

- **Schema**: `allowedModels: z.array(z.string()).optional()` already existed (`server/validators/appConfigSchema.js`).
- **Backend enforcement**: `RequestBuilder.filterModelsForApp()` already filters/falls back based on `app.allowedModels`.
- **Client-side filtering**: `client/src/utils/modelFiltering.js` and `ModelSelector.jsx` already filter the model dropdown by `app.allowedModels`.
- **Missing piece**: the admin App Editor had no field to set `allowedModels` — it could only be set by hand-editing the app's JSON config.

This PR closes that gap by adding an "Allowed Models" picker to the App Editor's basic settings (`AppFormEditor.jsx`), reusing the existing `ResourceSelector` component already used for the same purpose on Groups and OAuth clients. Leaving the selection empty preserves current behavior (no restriction, all models allowed).

Also added a changelog entry under `docs/releases/5.5.0/features.md` per project convention.

## Changes

- `client/src/features/admin/components/AppFormEditor.jsx`: import `ResourceSelector`, add `handleAllowedModelsChange`, render the picker next to "Preferred Model".
- `docs/releases/5.5.0/features.md`: changelog entry.

No backend or schema changes were needed — this is purely surfacing an existing config field in the admin UI.

## Test plan

- [x] `npx eslint` on the changed file — no new errors (pre-existing unrelated warnings only).
- [ ] Manually verify in the admin UI: open an app in the editor, add/remove models in "Allowed Models", save, confirm the chat model selector for that app is restricted accordingly.
- [ ] Confirm leaving "Allowed Models" empty keeps all models available (no regression for existing apps).

---
_Generated by [Claude Code](https://claude.ai/code/session_01R1TGgdwPwHV55rA25fF8WH)_

### DIFF
--- a/client/src/features/admin/components/AppFormEditor.jsx
+++ b/client/src/features/admin/components/AppFormEditor.jsx
@@ -6,6 +6,7 @@ import McpToolsSelector from './McpToolsSelector';
 import SkillsSelector from '../../../shared/components/SkillsSelector';
 import WorkflowsSelector from '../../../shared/components/WorkflowsSelector';
 import SourcePicker from './SourcePicker';
+import ResourceSelector from './ResourceSelector';
 import Icon from '../../../shared/components/Icon';
 import IconPicker from '../../../shared/components/IconPicker';
 import MimeTypeSelector from './MimeTypeSelector';
@@ -269,6 +270,14 @@ function AppFormEditor({
     onChange(updatedApp);
   };
 
+  const handleAllowedModelsChange = selectedModelIds => {
+    const updatedApp = {
+      ...app,
+      allowedModels: selectedModelIds
+    };
+    onChange(updatedApp);
+  };
+
   // Source handling functions
   const handleSourcesChange = selectedSourceIds => {
     const updatedApp = {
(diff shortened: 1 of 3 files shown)


manzke:
The `test-quick` check failure is unrelated to this PR's change (`AppFormEditor.jsx` / the Allowed Models picker). Diagnosis:

`tests/unit/server/locale-override.test.js` has two pre-existing bugs, uncovered because the `Unit Tests` GitHub Actions workflow was only added yesterday and appears to be running this suite in CI for the first time:

1. `import { ConfigCache } from '../configCache.js'` resolved to a nonexistent `tests/unit/configCache.js` (should be `server/configCache.js`).
2. `__dirname` was derived via `fileURLToPath(import.meta.url)`, but Jest transpiles this file to CommonJS via Babel, which can't parse raw `import.meta` — a hard syntax error, not a runtime issue.

I fixed both in `e2dc164` (isolated to this one test file, no shared config touched).

However, the suite **still fails** after that fix, for a deeper reason: with the import path corrected, `ConfigCache` now actually loads, and its dependency chain (`configLoader.js` → `server/pathUtils.js`) hits the *same* `import.meta.url` pattern — but this time in real production code (`pathUtils.js:6`), not a test file. That file is imported broadly across the server, so this isn't a one-file fix anymore.

Options to actually get `test-quick` green here:
- Add `babel-plugin-transform-import-meta` to the Jest transform in `tests/config/jest.config.js` (shims `import.meta.url` for Babel-transpiled test code only; doesn't touch production runtime, which runs as real ESM). This is the standard fix for this exact situation.
- Or quarantine/skip `locale-override.test.js` until that's decided.
- Or refactor `pathUtils.js` to avoid `import.meta` (touches production code used everywhere — higher risk, not recommended just to satisfy a test).

I didn't want to change shared Jest/Babel config or add a new dependency unilaterally from inside this unrelated feature PR, so I'm flagging it here rather than guessing. Let me know which direction you'd like and I'll follow up, or I'm happy to just proceed with the `babel-plugin-transform-import-meta` option if that sounds right.

---

PR #1881, opened 2026-07-07
refactor(websearch): resolve native search as a request-time directive, not a tool (adds Anthropic)

### LINKED ISSUE #739: Support native websearch and configuration for Anthrophic Claude Models

manzke:
## Summary

Closes #739.

Adds support for Anthropic's server-side [web search tool](https://platform.claude.com/docs/en/agents-and-tools/tool-use/web-search-tool) — and, per review feedback, redesigns how *all* native web search (Google, OpenAI, Anthropic) is resolved, since the original tool-file-based approach didn't scale cleanly to a third provider.

### Before → After

Previously, native search was modeled as a fake "tool" in `contents/tools/` (`googleSearch`, `webSearch`) with `isSpecialTool: true`. Every provider's generic tool-calling converter had to special-case-filter these ids and reconstitute them into the provider's native tool block. Adding Anthropic the same way meant a third copy of that pattern.

Now native search is a plain request-time directive, never a tool:

- `toolLoader.resolveNativeWebSearchProvider(modelProvider)` / `resolveAppNativeWebSearch(app, modelProvider, websearchEnabled)` return `{ provider }` or `null`.
- `RequestBuilder` (app chat) and `PromptNodeExecutor` (workflow agent nodes) resolve this and pass it through as a `nativeWebSearch` request option.
- `anthropic.js`, `google.js`, and `openai-responses.js` each inject their native tool block directly when building the request, independent of the regular tools array:
  - Anthropic: `{ type: 'web_search_20250305', name: 'web_search' }` — combinable with function tools.
  - OpenAI Responses: `{ type: 'web_search' }` — combinable with function tools.
  - Google: `{ google_search: {} }` — mutually exclusive with function calling (Gemini API limitation); function tools are dropped and a warning is logged.
- The generic tool-calling converters (`AnthropicConverter`, `GoogleConverter`, `OpenAIResponsesConverter`) go back to plain `provider`/`isSpecialTool` filtering, with zero special-case knowledge of native search.
- `googleSearch.json` and `webSearch.json` are removed from `contents/tools/` (migration `V069` cleans up existing installs). No `anthropicWebSearch.json` was ever added — native search never needs a tool file under this design.

### Response-side handling (unchanged in spirit, new for Anthropic)

`AnthropicConverter.convertAnthropicResponseToGeneric` handles the new response shapes for both streaming and non-streaming requests:
- `server_tool_use` blocks (the search query Claude issued) correctly produce no client-side tool call, since Anthropic executes the search itself.
- `web_search_tool_result` blocks (and their `web_search_tool_result_error` variant) are collected into `result.groundingMetadata.searchResults`.
- `citations_delta` streaming events and `citations` arrays on text blocks are collected into `result.groundingMetadata.citations`.
- `groundingMetadata` reuses the mechanism already used for Google Search grounding, so the "Grounding" answer-source badge lights up for Claude too.

### Workflow-level fix

Agent/prompt nodes previously had to hardcode `"tools": ["googleSearch"]` to get native grounding, and got **nothing** on Anthropic or OpenAI models. `PromptNodeExecutor` now resolves the generic `webSearch` marker through the same `resolveNativeWebSearchProvider`, falling back to the real `braveSearch` tool when the model has no native search. The three bundled research workflows (`iterative-research-human`, `iterative-research-auto`, `approval-workflow`) are updated to use `webSearch` instead of `googleSearch` (migration `V070` updates existing installs' copies, but only if untouched from the default). Citation capture (`_captureCitationsFromGroundingMetadata`) is extended to also handle Anthropic's grounding shape, so the workflow synthesizer can cite Anthropic web search results too.

### Known scope limits

Like the existing Google/OpenAI native search integrations, this doesn't persist raw provider-native content blocks (e.g. `encrypted_content`) across conversation turns — history is stored as flattened text, same as for every other provider. Anthropic's `pause_turn` continuation and "mixing a pending server-side search with a client tool call in the same turn" edge cases aren't specially handled.

## Test plan

- [x] `npm run lint:fix` / prettier on all changed files — no new errors, only pre-existing unrelated warnings
- [x] New tests in `server/tests/native-web-search.test.js`: `resolveNativeWebSearchProvider`/`resolveAppNativeWebSearch` unit tests, `anthropic.js`/`google.js`/`openai-responses.js` request-building tests (native tool injection, combining with function tools, Google's exclusivity + warning), and the Anthropic response-parsing tests (`server_tool_use`, `web_search_tool_result` success/error, `citations_delta`, full non-streaming response)
- [x] Updated `provider-specific-tool-filtering.test.js`, `openai-websearch-filtering.test.js`, `googleSearchFunctionCallingConflict.test.js` to reflect the new architecture (all pass)
- [x] Verified no regressions in `anthropic-tool-name.test.js`, `googleAdapter.test.js`, `google-tool-schema-validation.test.js`, `gemini3-*.test.js`, `agent-citation-contract.test.js` (confirmed several pre-existing, unrelated test failures in this sandbox via `git stash` bisection — not caused by this change)
- [ ] Manual smoke test against the live Anthropic API with a real API key (not available in this environment)

https://claude.ai/code/session_01Ueq1o2aaScsTp8wcDVmCDJ


manzke (commented):
@claude please check for me once. when native search is enabled, it will be picked for the models they support it.
for the models which do not, we use brave.
if native search is disabled, all models use brave.

manzke:
Confirmed — that's exactly what `resolveWebsearchTool` in `server/toolLoader.js` does today (unaffected by the tool-registration refactor we're still discussing above):

```js
let toolId;
if (useNativeSearch && modelProvider === 'google') {
  toolId = 'googleSearch';
} else if (useNativeSearch && modelProvider === 'openai-responses') {
  toolId = 'webSearch';
} else if (useNativeSearch && modelProvider === 'anthropic') {
  toolId = 'anthropicWebSearch';
} else {
  toolId = 'braveSearch';
}
```

- `useNativeSearch` is `true` by default (per-app override).
- **Native search enabled + model supports it** (Gemini/GPT/Claude) → the matching native tool (`googleSearch` / `webSearch` / `anthropicWebSearch`).
- **Native search enabled + model doesn't support it** (any other provider) → falls through to `braveSearch`.
- **Native search disabled** (`useNativeSearch: false`) → always `braveSearch`, regardless of provider.

So the behavior you described is already correct and covered — this PR only adds Anthropic to the "models that support it" list, it doesn't change the enabled/disabled or fallback logic.

Separately, I flagged above that removing the tool-registration approach entirely (as opposed to just adding Anthropic to it) would need to account for `googleSearch` also being referenced directly by ID in 3 default workflows — still waiting on your call on scoping that part.

---

manzke:
@claude does it make sense to keep the special tool ids or should we merge them into one like "nativeWebSearch"
if we decide against "nativeWebSearch", we should change "webSearch" to "openaiWebSearch"

manzke:
> does it make sense to keep the special tool ids or should we merge them into one like "nativeWebSearch"? if we decide against "nativeWebSearch", we should change "webSearch" to "openaiWebSearch"

They're already merged into one — the request-time-directive refactor on this branch removed every per-provider tool id (`googleSearch`, the OpenAI-specific `webSearch`, and `anthropicWebSearch` never existed as a file). The only id left is the generic **`webSearch` marker** on workflow nodes, which means "give this node web search": it resolves to the provider's native search on Google/OpenAI/Anthropic models and falls back to the real `braveSearch` tool everywhere else.

My recommendation is to keep that marker named `webSearch` rather than `nativeWebSearch`:
- `nativeWebSearch` would be misleading — the marker deliberately falls back to `braveSearch` on providers without native search, so it isn't always native.
- `openaiWebSearch` would reintroduce exactly the per-provider special-casing this refactor removed.
- Internally, the name `nativeWebSearch` **is** used where it's accurate: the adapter request option that's only set when provider-native search actually applies.
- Bonus: existing installs with `webSearch` in workflow configs keep working with zero migration.

This branch had also drifted into conflict with `main` (migration numbers V069/V070 are now taken, plus release-notes and #1906 overlap), so I've finalized everything in **#1930**: this branch rebased onto current `main`, migrations renumbered to V073/V074, plus three review fixes — an Anthropic citations-accumulation bug in workflow streaming (most citations were silently dropped), the V074 migration widened to cover customized workflow nodes still referencing `googleSearch`, and the bundled researcher system prompts that still told the model to call a `googleSearch` tool. Suggest closing this PR in favor of #1930.

---

manzke:
Closing as superseded by #1930, which merged this branch's content rebased onto `main` on 2026-07-09 (migrations renumbered to V073/V074, plus the workflow-streaming citation accumulation fix and the hardened V074 migration). #739 is closed.

Re-verified today against current `main`:
- Every code addition from this PR is present on `main`. The only differences are intentional: V074 was rewritten to cover customized workflow nodes, and the `nativeWebSearch` workflow adapter option now lives in the unified agentic loop's contract (#2261) instead of `adapterOptions.js`.
- `native-web-search.test.js`, `workflow-native-websearch-threading.test.js`, `provider-specific-tool-filtering.test.js`, `openai-websearch-filtering.test.js` and `googleSearchFunctionCallingConflict.test.js` all pass on `main`.

This branch is in conflict with `main` and has nothing left to merge.

---

PR #1900, opened 2026-07-08
fix(security): validate tool.script against path traversal

### LINKED ISSUE #1806: tool.script from config is joined into filesystem paths unvalidated — traversal-based arbitrary file read/write/delete
**Category:** security

### Problem
GET/PUT /api/admin/tools/:toolId/script and DELETE /api/admin/tools/:toolId build `join(rootDir, 'server', 'tools', tool.script)` from the stored `script` field without resolveAndValidatePath. POST/PUT tool routes validate only `id` and `name`, so an admin (or tampered tools.json) can set `script: '../../contents/.encryption-key'` or '../../server/server.js' and then read, overwrite (PUT with arbitrary content), or delete (DELETE unlinks the script) any file on disk relative to server/tools. This violates the project's own api-security rule ('Validate stored config paths too') that skills.js and pages.js already follow.

### Evidence
tools.js:948 `const scriptPath = join(rootDir, 'server', 'tools', tool.script); ... await fs.writeFile(scriptPath, content, 'utf-8');` — no path validation. tools.js:643-647 unlinks the same unvalidated join. tools.js:400 create/update validation: `if (!updatedTool.id || !updatedTool.name)` only.

### Recommended fix
Run tool.script through resolveAndValidatePath against server/tools (rejecting on null) in all three handlers, and validate the script field on create/update (e.g. /^[a-zA-Z0-9_-]+\.js$/).

### Affected files
- `server/routes/admin/tools.js:644`
- `server/routes/admin/tools.js:843`
- `server/routes/admin/tools.js:948`
- `server/routes/admin/tools.js:508`

---

manzke:
## Summary

Fixes #1806. The admin Tools API joined the stored `tool.script` filename directly into `server/tools/` without validating it, so a crafted `script` value (or a hand-edited `tools.json`) could be used to read, overwrite, or delete arbitrary files on disk via the GET/PUT `/api/admin/tools/:toolId/script` and DELETE `/api/admin/tools/:toolId` endpoints — e.g. `script: "../../contents/.encryption-key"`.

- `resolveAndValidatePath` (already used by `skills.js`/`pages.js`) is now applied to `tool.script` in all three unvalidated join sites: the GET script handler, the PUT script handler, and the DELETE tool handler's script cleanup. A path that resolves outside `server/tools/` is rejected (`400` for GET/PUT; skipped with a warning log for DELETE, consistent with existing script-cleanup error handling).
- The POST create and PUT update tool endpoints now reject a `script` value that isn't a bare `<name>.js` filename (`/^[a-zA-Z0-9_-]+\.js$/`), matching every script filename currently shipped in the repo.

## Test plan

- [x] Added `server/tests/admin-tools-script-path.test.js` covering: GET/PUT script traversal rejection, a legitimate script still working, DELETE skipping cleanup for a traversal path (tool config itself still deletes), and POST/PUT rejecting malformed script filenames.
- [x] `npm test -- tests/admin-tools-script-path.test.js` passes (6/6).
- [x] `npx eslint server/routes/admin/tools.js server/tests/admin-tools-script-path.test.js` clean.
- [x] Verified the same pre-existing test failures (`group-handling.test.js`, `admin-endpoints-security.test.js`) occur identically on `main`, unrelated to this change.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

---
_Generated by [Claude Code](https://claude.ai/code/session_01GfraQEYSpDv12p9esxMuWe)_


server/routes/admin/tools.js
```diff
@@ -892,7 +917,11 @@ export default function registerAdminToolsRoutes(app) {
       }
 
       const rootDir = getRootDir();
-      const scriptPath = join(rootDir, 'server', 'tools', tool.script);
+      const scriptsBaseDir = join(rootDir, 'server', 'tools');
+      const scriptPath = await resolveAndValidatePath(tool.script, scriptsBaseDir);
+      if (!scriptPath) {
+        return sendBadRequest(res, 'Invalid script path');
```
manzke: @claude that's a security issue. we should not tell the attacker if it is an invalid path. this allows enumeration. we should send a "not found" like in the other cases and log it.

manzke: Good catch — fixed in 7ff8bf8. Both the GET and PUT script handlers now return the same 404 "Script file" response for a rejected traversal path as they do for a genuinely missing file, so the two cases are indistinguishable to the caller. The real reason (out-of-bounds path) is now logged server-side via `logger.warn` instead.

---

PR #1905, opened 2026-07-08
Fix tool-argument repair corrupting valid JSON containing "}{" (#1802)

### LINKED ISSUE #1802: Tool-argument 'repair' replace(/}{/g, ',') corrupts valid JSON arguments before first parse
**Category:** bug

### Problem
executeToolCall mutates the raw tool-call arguments with `toolCall.function.arguments.replace(/}{/g, ',')` BEFORE attempting the first JSON.parse. This is meant to patch concatenated streaming fragments (`{...}{...}`), but it runs unconditionally, so perfectly valid JSON whose string values contain `}{` (e.g. a code snippet, regex, or text argument) is corrupted — `{"text":"}{"}` becomes `{"text":","}` (silently wrong args) or unparseable input that degrades to `args = {}`, making the tool run with empty arguments and no error surfaced to the model.

### Evidence
ToolExecutor.js:366-368: `let finalArgs = toolCall.function.arguments.replace(/}{/g, ',');` followed by `args = JSON.parse(finalArgs);` — the untouched original string is never tried first.

### Recommended fix
Try JSON.parse on the raw string first; only apply the `}{` merge heuristic as a fallback after a parse failure. Better: fix fragment merging where the chunks are accumulated (the 'smart concatenation' block) instead of post-hoc string surgery.

### Affected files
- `server/services/chat/ToolExecutor.js:366`

---

manzke:
## Summary

Fixes #1802.

`ToolExecutor.executeToolCall` applied the `}{` → `,` streaming-fragment repair heuristic
**unconditionally, before ever attempting to parse the raw tool-call arguments string**. Any
perfectly valid JSON whose string values happened to contain the literal substring `}{` (e.g. a
code snippet, regex, or text argument) was silently corrupted, or the corrupted string failed to
parse and the tool ran with empty arguments — with no error surfaced to the model.

- `JSON.parse` is now attempted on the raw, untouched arguments string first.
- Only when that fails do we fall back to the existing repair heuristics, in order: the `}{` → `,`
  merge (for genuinely concatenated streaming fragments like `{...}{...}`), then the
  bracket-wrapping repair, then giving up to `args = {}` with an error logged (unchanged final
  fallback behavior).
- Added `tests/unit/server/tool-executor-argument-parsing.test.js` covering: valid JSON containing
  a literal `}{` in a string value, concatenated streaming fragments still repairing correctly, and
  unparseable input still falling back to empty args.

## Test plan

- [x] `npm run test:unit -- tests/unit/server/tool-executor-argument-parsing.test.js` — new tests
      pass (185/185 server+client suites that could run in this sandbox passed; the 5 unrelated
      failing suites are pre-existing missing client `node_modules` in this environment, not
      caused by this change).
- [x] `npx eslint` / `npx prettier --check` on changed files — clean.

https://claude.ai/code/session_01L2Zzny3dtdhepabFM7vsbG

---
_Generated by [Claude Code](https://claude.ai/code/session_01L2Zzny3dtdhepabFM7vsbG)_

### DIFF
--- a/docs/releases/5.5.0/features.md
+++ b/docs/releases/5.5.0/features.md
@@ -207,6 +207,18 @@ resolve into a sibling directory that merely shared the skill's directory name a
 
 - No admin action required; existing skill packages install exactly as before.
 
+## Tool Calls With Certain Text Arguments No Longer Get Corrupted
+
+Tool-call arguments containing the literal characters `}{` in a text value (for example a code
+snippet, regex, or JSON example) are no longer corrupted before being parsed. A parsing safeguard
+meant to repair concatenated streaming fragments was previously applied to every tool call
+unconditionally, silently mangling otherwise valid arguments or causing the tool to run with no
+arguments at all.
+
+- Tool arguments are now parsed as-is first; the streaming-fragment repair only runs as a fallback
+  when the original arguments aren't valid JSON.
+- No configuration or admin action required.
+
 ## Chat No Longer Crashes When a Response Finishes
 
 Chat responses now complete cleanly instead of failing with an "Add-in Error" (`setSearchStatus is
--- a/server/services/chat/ToolExecutor.js
+++ b/server/services/chat/ToolExecutor.js
@@ -382,6 +382,10 @@ class ToolExecutor {
     let args = {};
 
     try {
+      args = JSON.parse(toolCall.function.arguments);
+    } catch {
+      // Raw string wasn't valid JSON — attempt repairs for known streaming artifacts
+      // (concatenated fragments like `{...}{...}`, or a truncated leading/trailing brace).
       let finalArgs = toolCall.function.arguments.replace(/}{/g, ',');
       try {
         args = JSON.parse(finalArgs);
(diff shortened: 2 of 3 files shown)


manzke:

**Verdict: superseded — recommend closing. Also a duplicate of intrafind/ihub-apps#1907 (opened 57 minutes later, same issue, same fix).**

- **Target file no longer exists**: `server/services/chat/ToolExecutor.js` isn't on `main` anymore — there's nothing left to apply this diff to.
- **The tool-calling pipeline was rewritten since**: the logic now lives in `server/services/loop/toolCallMerge.js`, whose own doc comment says it unifies rules "previously duplicated in... the former chat ToolExecutor and the workflow LLM helper." This is a from-scratch consolidation, not a patch on the old file.
- **The new code sidesteps this bug class entirely, more fundamentally than this PR's fix does**: streamed argument fragments are reassembled by plain string concatenation as they arrive (`existing.function.arguments += incomingArgs`) — there's no `}{` → `,` regex repair step anywhere in `server/` anymore, so a literal `}{` inside a valid string value is never at risk of being "repaired." The final `parseToolCallArguments()` does one `JSON.parse`, and on failure returns `{ __raw_arguments: raw }` instead of silently defaulting to `{}` — surfacing the problem to the model, which is strictly better than this PR's (and #1907's) preserved fallback of empty args + a server-only log.
- **Mergeability**: `mergeable_state` would be moot given the file is gone; this isn't a rebase situation, it's a rewrite-around.

**Recommendation**: close both this PR and #1907 as superseded by the `services/loop/toolCallMerge.js` rewrite — the underlying bug (#1802) already can't happen with the current architecture.

---

PR #1920, opened 2026-07-09
fix: delete prefill param alongside send after auto-send fires

### LINKED ISSUE #1916: Prefill autosend doesn't work
**Used Documentation**: https://github.com/intrafind/ihub-apps/blob/main/docs/auto-send-feature.md

**Description**: Opening a chat with query "?prefill=Welche%20quellen%20kennst%20du?&send=true" reloads a few times, then transforms the query into "?prefill=Welche+quellen+kennst+du%3F"

**Issue**: While the given prompt is correctly being prefilled, it it not being autosent.

Higher priority, a customer has complained about this.

manzke:
## Summary

Fixes #1916.

The auto-send effect in `AppChat.jsx` (used by the documented `?prefill={message}&send=true` "answer link" feature) only removed the `send` query parameter from the URL once auto-send fired, leaving `prefill` in place. A later reload of that URL (for example as part of an OIDC/SSO redirect round trip, which several full-page navigations are visible in for this feature) re-seeded `useState(prefillMessage)` and repopulated the chat input with the already-sent message — with `send` gone, nothing auto-sent again, so the user was left staring at text that looked unsent even though it had already gone out once. This matches the URL transformation (`?prefill=...&send=true` → `?prefill=...`) reported in the issue.

- `client/src/features/apps/pages/AppChat.jsx`: the auto-send effect now deletes both `send` and `prefill` from the URL, matching the existing settings-apply effect elsewhere in the same file that already cleans up both params together.
- `docs/auto-send-feature.md`: updated the troubleshooting note to reflect that both parameters are removed.
- `docs/releases/5.5.0/features.md`: added a changelog entry.

## Test plan

- [ ] Load an app with `?prefill=<message>&send=true`, confirm the message auto-sends and the URL settles with neither `prefill` nor `send` present.
- [ ] Reload the resulting (cleaned) URL and confirm the input box stays empty instead of repopulating with the old message.
- [ ] Combine with other params (e.g. `&model=...`) per `docs/auto-send-feature.md` examples and confirm no double-fire/race with the settings-apply effect.

https://claude.ai/code/session_0146CBWFPFMn1kn5VaHGRTkb

---
_Generated by [Claude Code](https://claude.ai/code/session_0146CBWFPFMn1kn5VaHGRTkb)_

### DIFF
--- a/client/src/auth-gate/auth-gate.js
+++ b/client/src/auth-gate/auth-gate.js
@@ -884,9 +884,10 @@
   // Resolve the returnUrl to forward to OIDC/NTLM. If the current URL already
   // carries a returnUrl query parameter (e.g. from /api/oauth/authorize
   // redirecting to /login?returnUrl=...), preserve it so the post-auth flow
-  // lands back on the original request. Falls back to the current URL minus
-  // its query string. Same-origin only to avoid open-redirect via the auth
-  // provider.
+  // lands back on the original request. Otherwise falls back to the current
+  // URL including its query string, so params like ?prefill=...&send=true
+  // (auto-send links) survive an OIDC/NTLM round trip for logged-out users.
+  // Same-origin only to avoid open-redirect via the auth provider.
   function getEffectiveReturnUrl() {
     try {
       var params = new URLSearchParams(window.location.search);
@@ -904,7 +905,7 @@
     } catch (e) {
       /* ignore */
     }
-    return window.location.href.split('?')[0];
+    return window.location.href;
   }
 
   // =========================================================================
--- a/client/src/features/apps/pages/AppChat.jsx
+++ b/client/src/features/apps/pages/AppChat.jsx
@@ -483,9 +483,11 @@ function AppChat({ preloadedApp = null }) {
     if (shouldAutoSend && !autoSendTriggered.current && prefillMessage && app && !processing) {
       autoSendTriggered.current = true;
 
-      // Clean up the send parameter from URL
+      // Clean up the send and prefill parameters from URL so a later reload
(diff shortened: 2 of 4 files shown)


manzke:
@claude If I'm logged in via OIDC, opening https://ihub.local.intrafind.io/apps/web-chat?prefill=Welche+quellen+kennst+du%3F&send=true works. Yes prefill is not deleted, but the chat is triggered.
With OIDC enabled (and logged out), opening https://ihub.local.intrafind.io/apps/web-chat?prefill=Welche+quellen+kennst+du%3F&send=true leads to an empty chat and no query parameters.

manzke:
Thanks for confirming both scenarios — pushed a second fix for the logged-out case.

Root cause: the pre-React `auth-gate.js` (which does the SSO auto-redirect before the SPA even mounts) resolves its OIDC/NTLM `returnUrl` via `getEffectiveReturnUrl()`. When the URL doesn't already carry a `returnUrl` param (the normal case for `?prefill=...&send=true`), it fell back to `window.location.href.split('?')[0]` — dropping the entire query string, including `prefill`/`send`, before storing it server-side for the post-login redirect. So logged-out + auto-redirect landed back on a bare URL with no params at all, matching "empty chat, no query parameters."

Changed the fallback to `window.location.href` (matching `AuthContext.jsx`'s `loginWithOidc`, which already defaults to the full href). `prefill`/`send` now survive the OIDC/NTLM round trip, and the existing token/provider cleanup (in both `auth-gate.js` and `AppChat.jsx`, per the first fix) removes only those params afterward.

Pushed as `1a16626` on the same branch/PR.

---

Issue #1788, opened 2026-07-06
server/tests mixes three incompatible test runners — `npm test` in server/ fails on 83 of 118 files


manzke:
**Severity:** medium · **Confidence:** suggestion · **Effort:** medium
**Category:** testing

### Problem
server/package.json's `test` script runs Jest with testMatch `**/tests/**/*.test.js`, but 83 of the 118 *.test.js files there are not Jest suites: most are plain node scripts with `assert`/console checks meant to be run as `node server/tests/x.test.js` (e.g. openaiAdapter.test.js, all agent-*.test.js), and 10 use `node:test` (adminRescue.test.js imports describe/it from 'node:test'). Jest fails each with 'must contain at least one test'. Root and server also pin different Jest majors (30.4.2 vs 29.7.0). Three runner conventions make the suite unrunnable as a whole.

### Evidence
server/jest.config.js:14 `testMatch: ['**/tests/**/*.test.js', ...]`; server/tests/openaiAdapter.test.js:1-17 is a top-level `assert.deepStrictEqual(...)` script with no describe/it; server/tests/agent-budget-loop.test.js:10 says 'Run directly: node server/tests/agent-budget-loop.test.js'; server/tests/adminRescue.test.js:1 `import { describe, it, ... } from 'node:test'`. Count: 83/118 files lack any top-level describe/it/test call. package.json:156 jest ^30.4.2 vs server/package.json:75 jest ^29.7.0.

### Recommended fix
Pick one runner (node:test needs no deps and already fits 10 files plus all plain scripts with minor wrapping). Rename non-conforming files (.test.js → .manual.js) or convert them, and align the Jest major if Jest stays.

### Affected files
- `server/jest.config.js:14`
- `server/package.json:10`
- `server/tests/openaiAdapter.test.js:1`
- `server/tests/adminRescue.test.js:1`

---
<sub>From the 2026-07-06 comprehensive review · tracking #1677 · report `concepts/2026-07-06 Comprehensive Code Review.md`</sub>

manzke:
## Implementation Plan

**What:** `npm test` in `server/` is broken because the Jest `testMatch` (`**/tests/**/*.test.js`) picks up every `*.test.js` file, but most of those files aren't Jest suites — they're plain `node script.js`-style scripts or `node:test` suites — so Jest fails them with "must contain at least one test." Fix: make every file matching that glob actually runnable by the configured runner, or repartition the files so each glob only matches its own runner's files.

[...]

PR #1934, opened 2026-07-09
refactor(config): consolidate configCache per-type loaders into a declarative registry

### LINKED ISSUE #1785: configCache: triplicated per-type load logic, inconsistent return shapes, and dead/broken getWithFallback
**Category:** simplification

### Problem
The 1715-line configCache implements the same per-type special-casing three times: initialize() (327-460), refreshCacheEntry() (579-751), and seven refreshXCache() methods (1192-1340) — ~350 lines of drift-prone duplication (refreshModelsCache skips the ETag comparison refreshCacheEntry does; credentials decryption lives in two places). Getter return shapes are inconsistent: getApps/getTools/getPrompts return a bare [] on cache miss but {data, etag} otherwise, forcing defensive `|| {}` at call sites (requestThrottler.js:22-23). getWithFallback (785-802) is both dead (zero callers) and broken — get() never returns null, so the file fallback is unreachable.

### Evidence
Nine special-case if-blocks duplicated between initialize (configCache.js:330-455) and refreshCacheEntry (:584-731); refreshAppsCache :1234-1249 duplicates the apps branch again without the ETag check. getApps miss path :853-854 `return [];` vs hit path :861-864 `return { data..., etag... }`. getWithFallback :786-789 `const cached = this.get(configPath); if (cached !== null) { return cached; }` with get() :758-763 always returning an object.

### Recommended fix
Introduce a declarative loader registry ({key, load(includeDisabled), postProcess}) consumed by initialize, refreshCacheEntry and a single generic refresh(key); normalize all getters to {data, etag}; delete getWithFallback.

### Affected files
- `server/configCache.js:327`
- `server/configCache.js:579`
- `server/configCache.js:1192`
- `server/configCache.js:785`
- `server/configCache.js:853`

---

manzke:
## Summary

Fixes #1785.

`server/configCache.js` special-cased the same seven-plus config types (apps, models, prompts, workflows, agents, tools, groups, platform, credentials) three separate times — in `initialize()`, `refreshCacheEntry()`, and the individual `refreshXCache()` methods — ~350 lines of drift-prone duplication. This PR consolidates that into a single declarative loader registry.

- **New `CONFIG_LOADERS` registry**: one entry per config type describing `load({ verbose, useCache })`, an optional `count(data)` for non-array data (groups/credentials), and an optional `onLoadErrorEmptyValue` for best-effort loaders (agent profiles). `initialize()` and `refreshCacheEntry()` both consume this registry instead of nine sequential `if (configPath === ...)` blocks each.
- **Fixes a real bug**: the six `refreshXCache()` wrappers (`refreshAppsCache`, `refreshModelsCache`, etc.) called their loader and unconditionally overwrote the cache entry, skipping the ETag-compare-before-write that `refreshCacheEntry()` already does. That meant every admin-triggered refresh bumped the cache's ETag — and invalidated downstream client caches — even when the reloaded data was byte-for-byte identical. All six now delegate to `refreshCacheEntry()`, so they get the same ETag comparison as the periodic TTL-driven refresh path.
- **Normalizes getter return shapes**: `getApps()`, `getTools()`, and `getPrompts()` returned a bare `[]` on cache miss but `{ data, etag }` on a hit, forcing defensive `|| {}` / `Array.isArray(...)` checks at call sites. They now always return `{ data, etag }`, matching `getModels()`/`getAgentProfiles()`/`getWorkflows()`.
- **Deletes dead code**: `getWithFallback()` was unreachable (`get()` never returns `null`, so its fallback branch could never run) and had zero real callers.
- **Fixes a live bug this surfaced**: `server/utils/sourceDependencyTracker.js` called `configCache.getApps() || []` and iterated the result directly as an array. Since `getApps()` actually returns `{ data, etag }` whenever the cache is populated (the normal case), `for (const app of apps)` was throwing on every call and being silently swallowed by the surrounding `try/catch` — so `getSourceUsage`/`getAllSourceDependencies`/`findOrphanedSources` were effectively always returning empty results. Fixed all three call sites to destructure `.data`.
- Dropped the now-redundant `|| {}` around `configCache.getTools()` in `server/requestThrottler.js`, since the return shape is guaranteed.

## Verification

- `npx eslint` / `npx prettier --check` clean on all changed files.
- Booted the dev server (`node server/server.js`) and confirmed every config type logs its expected "Cached X" message with the correct count, with no new errors.
- Wrote a standalone script exercising `configCache.initialize()` → `refreshAppsCache()` with no underlying file change, confirming the cache ETag is now stable across a no-op refresh (previously it would have changed). Also confirmed `getApps()` returns `{ data: [], etag: null }` on a simulated cache miss.
- Ran the existing test suites that exercise these code paths (`admin-tools-script-path.test.js`, `generalRoutes-appLanguage.test.js`, `nda-risk-analyzer.test.js`) — all pass; the one pre-existing failure (`admin-endpoints-security.test.js`, a `jest.mock()`-under-ESM issue tracked separately in #1788) reproduces identically on unmodified `main`, confirming it's unrelated to this change.

## Test plan

- [x] Lint/format clean
- [x] Dev server boots and all config types cache successfully
- [x] No-op refresh leaves the cache ETag unchanged (bug fix verified)
- [x] `getApps()`/`getTools()`/`getPrompts()` return `{ data, etag }` on both hit and miss
- [x] Relevant existing tests pass (pre-existing unrelated failure confirmed on `main`)

https://claude.ai/code/session_01LKRef3FBjvGpoU964BTd6D

---
_Generated by [Claude Code](https://claude.ai/code/session_01LKRef3FBjvGpoU964BTd6D)_

### DIFF
--- a/server/configCache.js
+++ b/server/configCache.js
@@ -268,6 +268,94 @@ function expandToolFunctions(tools = []) {
 }
 
 /**
+ * Declarative registry for the config types that need more than a plain
+ * `loadJson()` — special-cased loading, defaults, or post-processing.
+ * Consumed by `initialize()`, `refreshCacheEntry()`, and the `refreshXCache()`
+ * wrappers so each type's load/transform logic lives in exactly one place.
+ *
+ * `load({ verbose, useCache })` returns the resolved data, or `null` if the
+ * underlying file is missing/unreadable (matching `loadJson`'s contract).
+ * `count(data)` overrides the logged item count for non-array data; when
+ * omitted, `data.length` is used for arrays and no count is logged otherwise.
+ * `onLoadErrorEmptyValue` marks loaders whose failures should be swallowed
+ * (logged as a warning, cache set to this fallback value) instead of
+ * propagating — matching agent profiles' existing "best effort" loading.
+ */
+const CONFIG_LOADERS = {
+  'config/apps.json': {
+    label: 'apps',
+    load: ({ verbose }) => loadAllApps(true, verbose)
+  },
+  'config/models.json': {
+    label: 'models',
+    load: ({ verbose }) => loadAllModels(true, verbose)
+  },
+  'config/prompts.json': {
+    label: 'prompts',
+    load: ({ verbose }) => loadAllPrompts(true, verbose)
+  },
+  'config/workflows.json': {
+    label: 'workflows',
+    load: ({ verbose }) => loadAllWorkflows(true, verbose)
+  },
+  'config/agents.json': {
+    label: 'agent profiles',
+    load: ({ verbose }) => loadAllAgentProfiles(true, verbose),
+    onLoadErrorEmptyValue: []
+  },
+  'config/tools.json': {
+    label: 'tools',
+    load: async ({ verbose }) => expandToolFunctions(await loadAllTools(true, verbose))
+  },
+  'config/groups.json': {
+    label: 'groups with resolved inheritance',
+    load: async ({ useCache }) => {
+      const groupsConfig = await loadJson('config/groups.json', { useCache });
+      return groupsConfig !== null ? resolveGroupInheritance(groupsConfig) : null;
+    },
+    count: data => Object.keys(data.groups || {}).length
+  },
+  'config/platform.json': {
+    label: 'platform config',
+    load: async ({ useCache }) => {
+      const platformData = await loadJson('config/platform.json', { useCache });
+      if (platformData === null) return null;
+      // Decrypt the realtime speech API key so the WS proxy receives
+      // plaintext. Env-var placeholders are resolved later in setCacheEntry.
+      if (platformData.speech?.realtime?.apiKey) {
+        platformData.speech.realtime.apiKey = decryptIfEncrypted(
+          platformData.speech.realtime.apiKey
+        );
+      }
+      // Decrypt the Azure Speech subscription key so the token broker
+      // (/api/voice/azure/token) can exchange it for a short-lived token.
+      if (platformData.speech?.azure?.subscriptionKey) {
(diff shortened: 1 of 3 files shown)


manzke:

**Verdict: still valid, and arguably more valuable now than when opened — the duplication it targets has grown, not shrunk.**

- **Both specific bugs confirmed present verbatim on `main`**:
  - `getApps()` still returns a bare `[]` on cache miss (`server/configCache.js:1111`) vs. `{ data, etag }` on a hit (~1119) — the exact inconsistency this PR normalizes.
  - `refreshAppsCache()` still calls `this.setCacheEntry('config/apps.json', apps)` directly (`configCache.js:1507`) instead of routing through `refreshCacheEntry()`'s ETag-compare-before-write — so a no-op admin-triggered refresh still bumps the ETag and invalidates downstream client caches today, exactly as described.
- **The underlying duplication has grown since**: `configCache.js` is now 2031 lines, and 4 more config types (`sources`, `providers`, `registries`, `installations`) have been added since this PR was written, each following the same `refreshCacheEntry()`-bypassing pattern this PR set out to consolidate. The declarative `CONFIG_LOADERS` registry this PR proposes doesn't cover these newer types, so landing this PR as-is would still leave some duplication — but the registry pattern would make adding them properly much easier than today's ad hoc approach.
- **Mergeability**: GitHub reports `mergeable_state: dirty` (base from 2026-07-09, ~2.5 months drift on a file that's clearly still actively grown).
- **No reviews/comments** — never looked at.

**Recommendation**: keep open and worth prioritizing — real, still-present bugs (not just style), and rebasing this now would make the next 4 config-type additions (and whatever comes after) cheaper too. Expect to fold `sources`/`providers`/`registries`/`installations` into the new registry while rebasing, not just resolve textual conflicts.

---

PR #1938, opened 2026-07-09
fix: apply Zod defaults via lenient fallback when strict schema rejects unknown keys

### LINKED ISSUE #1803: Strict app/model schema drops Zod defaults when config has any unknown key
**Severity:** low · **Confidence:** verified · **Effort:** medium
**Category:** bug

### Problem
createSchemaValidator applies parsed data (with Zod defaults + unknown-key stripping) only when safeParse succeeds; on failure it returns the ORIGINAL unparsed item. Because appConfigSchema and modelConfigSchema use .strict(), any config containing a single stray/legacy key makes safeParse fail — so the schema-provided defaults (e.g. enabled, sendChatHistory, preferredOutputFormat) are silently NOT applied and the raw object is used, while only a warning is logged. Configs that 'mostly work' can thus behave differently from what the schema documents.

### Evidence
resourceLoader.js:358-373 `const result = schema.safeParse(item); if (!result.success) { logger.warn(...) } else { validatedItem = result.data; }` — failure path keeps `item`. appConfigSchema.js:381 `.strict()`; modelConfigSchema.js:169 `.strict()`. A separate knownKeys warning (resourceLoader.js:377) confirms unknown keys are expected in the wild.

### Recommended fix
On parse failure, fall back to parsing against a non-strict variant (or `.passthrough()`) so defaults still apply, or surface the validation error instead of silently using the raw object. Decide explicitly whether strict should reject-load or lenient-load.

### Affected files
- `server/utils/resourceLoader.js:350`
- `server/validators/appConfigSchema.js:381`

---

manzke:
## Summary

- `createSchemaValidator` (`server/utils/resourceLoader.js`) only applied Zod's parsed output — including schema defaults like `enabled`, `sendChatHistory`, `preferredOutputFormat` — when `safeParse` succeeded. Because `appConfigSchema` and `modelConfigSchema` use `.strict()`, a single stray/legacy key on an otherwise-valid app or model config made `safeParse` fail entirely, so the loader silently fell back to the raw, unvalidated object with only a `logger.warn`, and defaults were never applied.
- Added `appConfigSchemaLenient` / `modelConfigSchemaLenient`: the same schemas (with the same refinements) but built with `.passthrough()` instead of `.strict()`, so unknown keys don't cause the whole parse to fail.
- `createSchemaValidator` now takes an optional third `lenientSchema` argument. When the strict parse fails, it retries against the lenient schema; if that succeeds, the defaults-applied/lenient-parsed data is used (and a distinguishing warning is logged) instead of the raw item. If the lenient parse also fails (i.e. the config is invalid for reasons other than unknown keys, like a missing required field), behavior is unchanged from before — raw item + warning.
- Wired the lenient schemas through `appsLoader.js` and `modelsLoader.js`.

This matches the recommended default from the issue's own open question: reject-vs-lenient-load is only changed for the "unknown key only" case; genuinely invalid configs keep today's behavior.

## Test plan

- [x] Added `server/tests/resourceLoader-lenient-schema.test.js` covering: defaults apply on a clean strict parse; defaults still apply via lenient fallback when an unrecognized key is present; raw item is still returned when the config is invalid for other reasons (missing required field); model schema gets the same treatment; and behavior is unchanged when no lenient schema is passed (agents/workflows loaders, unaffected).
- [x] `node --experimental-vm-modules node_modules/jest/bin/jest.js tests/resourceLoader-lenient-schema.test.js` — 5/5 passing.
- [x] `npm run lint:fix && npm run format:fix` on changed files — clean.
- [x] `node server.js` boots and loads apps/models without errors.

Fixes #1803

https://claude.ai/code/session_01Ev16aQNgfoj5Duc4uuF1A1

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ev16aQNgfoj5Duc4uuF1A1)_

### DIFF
--- a/server/appsLoader.js
+++ b/server/appsLoader.js
@@ -1,5 +1,9 @@
 import { createResourceLoader, createSchemaValidator } from './utils/resourceLoader.js';
-import { appConfigSchema, knownAppKeys } from './validators/appConfigSchema.js';
+import {
+  appConfigSchema,
+  appConfigSchemaLenient,
+  knownAppKeys
+} from './validators/appConfigSchema.js';
 
 /**
  * Enhanced Apps Loader Service
@@ -15,7 +19,7 @@ const appsLoader = createResourceLoader({
   resourceName: 'Apps',
   legacyPath: 'config/apps.json',
   individualPath: 'apps',
-  validateItem: createSchemaValidator(appConfigSchema, knownAppKeys)
+  validateItem: createSchemaValidator(appConfigSchema, knownAppKeys, appConfigSchemaLenient)
 });
 
 /**
--- a/server/modelsLoader.js
+++ b/server/modelsLoader.js
@@ -1,5 +1,9 @@
 import { createResourceLoader, createSchemaValidator } from './utils/resourceLoader.js';
-import { modelConfigSchema, knownModelKeys } from './validators/modelConfigSchema.js';
+import {
+  modelConfigSchema,
+  modelConfigSchemaLenient,
+  knownModelKeys
+} from './validators/modelConfigSchema.js';
 import logger from './utils/logger.js';
 
 /**
@@ -46,7 +50,7 @@ const modelsLoader = createResourceLoader({
   resourceName: 'Models',
   legacyPath: 'config/models.json',
   individualPath: 'models',
-  validateItem: createSchemaValidator(modelConfigSchema, knownModelKeys),
+  validateItem: createSchemaValidator(modelConfigSchema, knownModelKeys, modelConfigSchemaLenient),
   postProcess: ensureOneDefaultModel
 });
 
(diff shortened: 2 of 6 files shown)


manzke:

**Verdict: still valid, unimplemented, exactly as described.**

- **Bug confirmed present verbatim**: `createSchemaValidator` in `server/utils/resourceLoader.js:365` still only takes `(schema, knownKeys = [])` — no third `lenientSchema` parameter — and still does a single `schema.safeParse(item)` with no fallback retry (line 374).
- **No lenient variants exist**: `appConfigSchema.js` and `modelConfigSchema.js` both still end with `.strict()` (lines 446 and 237 respectively) — no `appConfigSchemaLenient`/`modelConfigSchemaLenient` anywhere. A single stray/legacy key on an otherwise-valid app or model config still silently skips all Zod defaults today, exactly as this PR diagnoses.
- **Not superseded**: nothing else has touched this validation path since.
- **Mergeability**: GitHub reports `mergeable_state: dirty` (base from 2026-07-09, ~2.5 months drift on a 6-file diff) — worth a normal rebase check but no architectural surprises expected given how self-contained the change is.
- **No reviews/comments** — never looked at.

**Recommendation**: keep open — rebase and merge; real, still-present bug with a well-scoped, tested fix.

---

PR #1988, opened 2026-07-11
fix: enforce user model permissions in chat flow (#1734)

### LINKED ISSUE #1734: Chat flow does not enforce user model-level permissions
**Severity:** medium · **Confidence:** verified · **Effort:** medium
**Category:** security

### Problem
prepareChatRequest resolves the model via filterModelsForApp, which filters only by app requirements (allowedModels, tool support, settings.model.filter) — never by the requesting user's permissions.models. The chat POST route also has no modelAccessRequired middleware. So a user (or anonymous principal) can pass any modelId the app permits and use it, ignoring their group's models allowlist. This is inconsistent with /api/models and /api/models/:modelId, which both filter/deny by user.permissions.models. A restricted group with apps:['*'] but models:['cheap-model'] can invoke expensive models via any app that allows them.

### Evidence
RequestBuilder.js:82-107 filterModelsForApp takes only (models, app). RequestBuilder.js:205 `let resolvedModelId = modelId || app.preferredModel || defaultModel;` then validates only against the app-filtered list (line 229). The `user` param is used for prompts/tools/completion but never for model permission filtering.

### Recommended fix
Intersect the app-filtered model list with user.permissions.models (honoring '*') before resolving/validating the requested model, and return 403 when the requested model is outside the user's allowance. If app-scoped model access is intended, document it and align the list endpoints.

### Affected files
- `server/services/chat/RequestBuilder.js:82`
- `server/services/chat/RequestBuilder.js:205`

---

manzke:
## Summary

Fixes #1734.

`RequestBuilder.prepareChatRequest` (the code path behind `POST /api/apps/:appId/chat/:chatId`) resolved and validated the chat model purely against app-level constraints (`allowedModels`, tool support, `settings.model.filter`) and never consulted `req.user.permissions.models`. A user in a group whose `models` allowlist excluded an expensive/restricted model could still invoke it through any app that permitted it — inconsistent with `/api/models` and the OpenAI-compatible proxy (`/api/inference/v1/*`), which both already enforce `permissions.models`.

- Added `filterModelsForUser`, which narrows the app-filtered model list down to the user's group-permitted models (reusing the existing `filterResourcesByPermissions` helper from `utils/authorization.js`).
- If the caller explicitly requests a `modelId` that exists but is outside the user's allowlist, the request now fails fast with a new `modelAccessDeniedForUser` error code (mapped to HTTP 403), instead of silently falling back to a different, permitted model.
- Default/fallback model resolution (no explicit `modelId`, app's `preferredModel`, global default, first-available fallback) now only considers the intersection of app-permitted and user-permitted models.
- If that intersection is empty, the existing `noModelsForUser` error path is used, matching prior behavior when nothing is resolvable.
- Added new `modelAccessDeniedForUser` i18n strings (en/de).

## Test plan

- [x] New regression test suite `server/tests/model-permission-enforcement.test.js` (Jest) covering:
  - explicit request for a model outside the user's allowlist → rejected with `modelAccessDeniedForUser`
  - default resolution never falls back to a model outside the permitted set
  - empty intersection between app-permitted and user-permitted models → `noModelsForUser`
  - wildcard (`*`) permission unaffected
  - requests without a `user`/`permissions` object (internal callers) unaffected
- [x] `npx eslint` and `npx prettier --check` on all changed files — clean
- [x] Ran the full test file with `node --experimental-vm-modules node_modules/jest/bin/jest.js tests/model-permission-enforcement.test.js` — all 5 tests pass

https://claude.ai/code/session_01S7vgTCxnS7ZyMrexReSPAV

---
_Generated by [Claude Code](https://claude.ai/code/session_01S7vgTCxnS7ZyMrexReSPAV)_

### DIFF
--- a/docs/releases/5.5.0/features.md
+++ b/docs/releases/5.5.0/features.md
@@ -475,3 +475,17 @@ is now both prevented and, if it still happens, reported clearly instead of show
   ("The AI model returned an incomplete response… please try sending your message again") rather
   than a silent blank reply.
 - No admin action is required — the fix takes effect automatically on upgrade.
+
+## Chat Requests Now Enforce Per-User Model Permissions
+
+The model allowlist configured for a user's group (**Admin → Groups → Models**) is now enforced on
+every chat request, closing a gap where it previously only affected which models appeared in the
+app's model picker.
+
+- Requesting a model outside your group's allowed models now returns a clear "not permitted" error
+  instead of silently succeeding.
+- Automatic model selection (when no model is explicitly chosen) now only considers models both the
+  app and the user's group permit, instead of every model the app allows.
+- No admin action is required — groups that already restrict `models` to specific IDs are now fully
+  enforced for chat, matching the behavior already used by the model list and the OpenAI-compatible
+  API endpoints.
--- a/server/routes/chat/sessionRoutes.js
+++ b/server/routes/chat/sessionRoutes.js
@@ -811,12 +811,14 @@ export default function registerSessionRoutes(
               .status(
                 prep.error.code === 'APP_NOT_FOUND' || prep.error.code === 'MODEL_NOT_FOUND'
                   ? 404
-                  : prep.error.code === 'noModelsAvailable' ||
-                      prep.error.code === 'noCompatibleModels' ||
-                      prep.error.code === 'noModelIdProvided' ||
-                      prep.error.code === 'noModelsForUser'
-                    ? 400
-                    : 500
+                  : prep.error.code === 'modelAccessDeniedForUser'
+                    ? 403
+                    : prep.error.code === 'noModelsAvailable' ||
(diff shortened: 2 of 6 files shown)


manzke:
Closing in favor of #2464, which finishes this fix rebased onto current `main`.

This PR predates a case-insensitive id-matching change (#2332) that `main` has since gained, and a partial fix (`isModelPermittedForUser`, covering only the explicit-`modelId` path) landed directly on `main` afterward — so this branch no longer applies cleanly and duplicates part of what's now already there. #2464 reconciles both: it keeps the case-insensitive matching, replaces the partial explicit-request check with a hard `modelAccessDeniedForUser` (403) as originally proposed here, and — the part that was still missing either way — bounds the *default/fallback* resolution path (app's `preferredModel`, global default, fallback chain) by the user's `permissions.models` too, not just explicit requests.

---

PR #1993, opened 2026-07-12
fix: stop InMemorySink actionTracker listener leak on non-tools app-as-tool calls

### LINKED ISSUE #1726: InMemorySink leaks a fire-sse listener on the actionTracker singleton for every non-tools App-as-tool invocation
**Severity:** medium · **Confidence:** verified · **Effort:** small
**Category:** bug

### Problem
invokeAppInternal always calls sink.startListening() (registers a listener on the shared actionTracker EventEmitter). For apps without tools, the response arrives via the res-shim (NonStreamingHandler writes res.json), and getResult takes the early-return branch `if (this.jsonBody && this.chunks.length === 0 && !this.done) return this._assembleNonStreamingResult();` — which never calls stopListening(). stopListening only runs on the streaming path or in invokeAppInternal's catch. Every successful non-tools app invocation therefore leaves a permanent listener holding the sink alive: max-listener warnings, growing heap, and dead-listener iteration on busy agent servers.

### Evidence
InMemorySink.js:162-171 — both early returns precede `this.stopListening()` at line 185; ChatService.js:308-320 non-tools path calls processNonStreamingChat then sink.getResult with no stopListening on success.

### Recommended fix
Call this.stopListening() at the top of getResult (or in a finally), and have invokeAppInternal stop the sink in a finally block instead of only in catch.

### Affected files
- `server/services/chat/streamSink/InMemorySink.js:164`
- `server/services/chat/ChatService.js:308`

---

manzke:
## Summary

Fixes #1726.

`InMemorySink.getResult()` had an early-return branch for the non-streaming (tool-less) App-as-tool success path that returned before reaching `this.stopListening()`. Since `actionTracker` is a process-wide singleton `EventEmitter`, every successful tool-less App-as-tool invocation left a `fire-sse` listener registered forever — growing heap (each closure retains the `InMemorySink` instance) and eventual `MaxListenersExceededWarning` noise under sustained agent traffic.

- `server/services/chat/streamSink/InMemorySink.js`: wrap `getResult()`'s body in `try/finally` so **every** exit path (both early-return JSON-body branches, normal streaming completion, and the timeout branch) calls `stopListening()`. This removes the need for the previous `this.stopListening()` call placed only after the streaming-wait, since it's now hoisted into the `finally`.
- `server/services/chat/ChatService.js`: `invokeAppInternal` now releases the sink in a `finally` block instead of only in the `catch`, so a throw before `sink.getResult()` is reached (e.g. from `prepareChatRequest`) is also covered. Removed the now-redundant explicit `sink.stopListening()` calls at the two return sites.
- Added `server/tests/inmemory-sink-listener-leak.test.js`, following the existing `inmemory-sink-overflow.test.js` pattern (plain Node script, no test framework), asserting `actionTracker.listenerCount('fire-sse')` returns to baseline after the non-streaming success path, the streaming success path, and the timeout path.

## Test plan

- [x] `node server/tests/inmemory-sink-listener-leak.test.js` — new regression test, all assertions pass
- [x] `node server/tests/inmemory-sink-overflow.test.js` — existing sink tests still pass (no regression)
- [x] `npx eslint` on the three changed files — clean

---
_Generated by [Claude Code](https://claude.ai/code/session_01PhHaFrHvkrD16DVX7uffBh)_

### DIFF
--- a/server/services/chat/ChatService.js
+++ b/server/services/chat/ChatService.js
@@ -84,7 +84,6 @@ class ChatService {
         variables
       });
       if (!prepResult.success) {
-        sink.stopListening();
         return {
           status: 'error',
           error: prepResult.error,
@@ -166,7 +165,6 @@ class ChatService {
       }
       return result;
     } catch (error) {
-      sink.stopListening();
       logger.error('invokeAppInternal failed', {
         component: 'ChatService',
         appId,
@@ -179,6 +177,11 @@ class ChatService {
         finalMessage: null,
         toolCalls: []
       };
+    } finally {
+      // Defense-in-depth: sink.getResult() already releases the listener on
+      // every exit path it controls, but this also covers callers that throw
+      // before getResult() is reached (e.g. prepareChatRequest/processChat*).
+      sink.stopListening();
     }
   }
 }
--- a/server/services/chat/streamSink/InMemorySink.js
+++ b/server/services/chat/streamSink/InMemorySink.js
@@ -160,49 +160,55 @@ export class InMemorySink {
    * Wait for the chat to finish and return the assembled result.
    */
   async getResult({ timeoutMs = DEFAULT_TIMEOUT_MS } = {}) {
-    // If we have a non-streaming JSON body (NonStreamingHandler path), return it directly.
-    if (this.jsonBody && !this.chatId) {
-      return this._assembleNonStreamingResult();
-    }
+    try {
+      // If we have a non-streaming JSON body (NonStreamingHandler path), return it directly.
+      if (this.jsonBody && !this.chatId) {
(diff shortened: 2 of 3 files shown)


manzke:
Superseded — the bug this fixes (#1726) describes code that no longer exists on `main`. `server/services/chat/streamSink/InMemorySink.js` was removed and `ChatService.invokeAppInternal` was rewritten around the newer `AgentLoop`/seam architecture, which never registers a listener on the shared `actionTracker` `EventEmitter` in the first place. This PR's diff no longer applies to current `main`.

Closing as stale/no longer applicable. See #1726 for the full verification.

---

PR #2030, opened 2026-07-13
Scope admin filesystem-source file API to contents/sources

### LINKED ISSUE #1687: Scope admin sources file API to contents/sources — contentAdmin can read/write all of contents/ (privilege escalation)
**Severity:** high · **Confidence:** evidence · **Effort:** small
**Category:** security

### Problem
The filesystem-source file endpoints (GET/POST/DELETE /api/admin/sources/:id/files*) are guarded by contentAdminAuth (weaker than adminAccess) and pass the raw request `path` to FileSystemHandler, whose basePath is the ENTIRE contents/ directory, not contents/sources. A contentAdmin can read `contents/.encryption-key` and `config/users.json` (bcrypt hashes), and write `config/groups.json` or `config/platform.json` — e.g. grant their own group `adminAccess: true` — escalating to full admin and decrypting every stored secret. The same root cause lets a filesystem source's `config.path` point at any file under contents/ (schema only blocks `..`, `~`, absolute paths).

### Evidence
sources.js:1891 `const result = await handler.writeFile(path, content, encoding);` with `path` from req.body, route guarded only by `contentAdminAuth` (sources.js:1862). FileSystemHandler.js:21 `this.basePath = handlerConfig.basePath || path.join(rootDir, contentsDir);` (whole contents/). sourceConfigSchema.js:217 only rejects `path.includes('..') || path.includes('~')`, so `config/groups.json` and `.encryption-key` are valid paths.

### Recommended fix
Give the admin file routes (and filesystem sources) a handler whose basePath is `contents/sources`, or validate the requested/stored path with resolveAndValidatePath against contents/sources. Additionally deny-list dotfiles (.encryption-key, .jwt-*) and config/.

### Affected files
- `server/routes/admin/sources.js:1859`
- `server/routes/admin/sources.js:1742`
- `server/routes/admin/sources.js:1972`
- `server/sources/FileSystemHandler.js:21`
- `server/validators/sourceConfigSchema.js:215`

---

manzke:
## Summary

Fixes #1687 (sev:high, privilege escalation). A `contentAdmin` user — a role deliberately weaker than full `adminAccess` — could use the filesystem-source browse/read/write/delete/test endpoints to reach any file under `contents/`, not just `contents/sources/`. That included `contents/.encryption-key` (used to decrypt all stored secrets) and `contents/config/groups.json` (writable to grant themselves `adminAccess: true`).

Root cause: `FileSystemHandler.basePath` is the whole `contents/` directory, and `_resolveSafePath()` only checked that a resolved path stayed inside `contents/`, never that it stayed inside `contents/sources/`. `sourceConfigSchema.js`'s `validateFilesystemPath()` had the same gap, and `SourceManager.testFilesystemSource()` re-implemented the same containment check independently.

- `server/sources/FileSystemHandler.js`: `_resolveSafePath()` now enforces a second containment check against the canonical `contents/sources` directory (cached via `_getSourcesDir()`) — this is the single method every read/write/list/delete/exists call funnels through, so it closes the gap for all four file routes plus runtime source loading in one place. `validateConfig()` also rejects paths outside `sources/` as a second layer.
- `server/sources/SourceManager.js`: `testFilesystemSource()` now additionally validates the resolved path stays within `contents/sources` (closing the same info-leak on the `/test` endpoint).
- `server/validators/sourceConfigSchema.js`: `validateFilesystemPath()` now rejects any filesystem source `config.path` that isn't `sources` or doesn't start with `sources/`, and rejects any path segment starting with `.` (blocks dotfiles/dot-directories).
- `server/routes/admin/sources.js`: the browse route's default `path` query param changed from `''` (which already threw, since `contents/` root itself is now out of bounds) to `'sources'`.

No data migration needed — every shipped default filesystem source already stores its path with the `sources/` prefix (`server/defaults/config/sources.json`), and the client already writes uploads as `sources/${fileName}` (`AdminSourceEditPage.jsx`, `FileUploader.jsx`).

## Test plan

- [x] New unit tests in `server/tests/fileSystemHandler-sources-scope.test.js`: reads/writes/deletes under `sources/` succeed; reads/writes/deletes targeting `config/groups.json` and `.encryption-key` (both still lexically inside `contents/`) are rejected; `validateConfig()` rejects out-of-scope paths.
- [x] Updated `server/tests/source-validation.test.js`: added cases rejecting `data/file.txt`, `config/groups.json`, and `sources/.env`; updated the pre-existing "minimal config" and "whitespace-only path" cases, which are now correctly rejected by the tightened schema validation.
- [x] `npx eslint` and `npx prettier --check` clean on all changed files.
- [x] Confirmed pre-existing, unrelated test failures (`admin-endpoints-security`, `adminRescue`, `auto-source-addition`, `office365-old-token-migration`) are present on `main` too — not a regression from this change.
- [x] Added a changelog entry to `docs/releases/5.5.0/features.md`.

https://claude.ai/code/session_014x3hitcEC5396icD3WmqV5

---
_Generated by [Claude Code](https://claude.ai/code/session_014x3hitcEC5396icD3WmqV5)_


manzke:
Thanks for the review — addressed all three points in 3fc96b6:

1. `SourceManager.testFilesystemSource` (and `FileSystemHandler._resolveSafePath`) no longer re-implement containment with a manual `startsWith` check. Both now strip the `sources/` prefix and delegate the actual boundary/symlink validation to `resolveAndValidatePath()` against the `contents/sources` directory itself — one canonical implementation.
2. `validateFilesystemPath()` now applies the `sources/`-prefix and dotfile-segment checks unconditionally, so the old `/app/`/`/workspace/` absolute-path carve-out can no longer bypass them (that carve-out never worked at the runtime-handler layer anyway, since `_resolveSafePath` requires the path to literally start with `sources/`). Added a regression test for an absolute path (`/app/config/groups.json`).
3. Split the changelog: the security fix stays in `features.md`, and a new `breaking-changes.md` entry covers the concrete admin action needed for any custom (non-bundled) filesystem source whose path lives outside `sources/`.

---

manzke:
Re-checked the three review threads against the code on `3fc96b6`:

- **Absolute-path bypass in `validateFilesystemPath()`** — fixed and verified; the `sources/` and dot-segment checks now run unconditionally, and `/app/config/groups.json` is rejected by test. Thread resolved.
- **Changelog upgrade note too broad** — fixed and verified; `breaking-changes.md` now names the concrete admin action for custom out-of-scope sources. Thread resolved.
- **Duplicate containment implementation in `SourceManager`** — only *half* addressed, so I've left this thread open. The containment check itself is centralized now, but the "strip the `sources/` prefix, then resolve" step is still written twice, and the two copies had already drifted in two ways:
  1. `testFilesystemSource()` skips `sanitizeRelativePath()`, so it rejects `/sources/faq.md` — a path the read/write handler accepts. "Test connection" fails on a source that loads fine.
  2. `testFilesystemSource()` resolves against a hardcoded `<root>/contents/sources`, ignoring `CONTENTS_DIR`, so on an installation with a custom contents directory it probes the wrong directory. (Pre-existing — the old code had it too.)

  #2220 collapses both call sites onto one public `FileSystemHandler.resolveSourcePath()`, which fixes both, and adds regression tests that fail on the current code. It targets this branch, so merging it folds the fix into this PR.

Also confirmed CI is green on `3fc96b6` and the full unit suite (430 tests) passes locally.

---

Issue #1497, opened 2026-05-20
[Chat] Resumable SSE streams — buffer + Last-Event-ID replay


manzke:
## Phase 2B — Resumable Streams

SSE stream buffering so network hiccups don't lose partial AI responses.

### Implementation
- In-memory stream buffer with configurable TTL (5 minutes default)
- On stream start, buffer chunks in `MemoryStreamBuffer` (`Map<string, Chunk[]>`)
- Each SSE chunk gets a monotonic index as `id:` field
- Client reconnects with `Last-Event-ID` header to resume from last received chunk
- For multi-instance deployments, completed messages persisted in PostgreSQL — reconnecting clients can restore from DB if buffer expired

### New Module: `server/services/streaming/`
```
StreamBufferService.js      — manages in-memory stream buffers
MemoryStreamBuffer.js       — Map<string, Chunk[]> with setTimeout TTL cleanup
```

### Files Modified
- `server/sse.js` — refactor: in-memory buffer, replay from offset
- `server/services/chat/StreamingHandler.js` — integrate buffer writes
- New: `client/src/features/chat/hooks/useResumableStream.js`

**Feature flag:** `resumableStreams` (default: true)

manzke:
## Implementation plan

**What:** SSE chat streams currently die completely on any network hiccup — the client has no reconnect logic at all, so a dropped connection mid-response means the partial answer is gone and the user must resend; this issue adds server-side chunk buffering with a monotonic event id and client-side reconnect-with-replay so a brief network blip doesn't lose an in-flight AI response.

**Why:** Chat is the core product surface and mobile/flaky-wifi users hit this regularly; today a dropped socket is indistinguishable from a hard failure (issue #1497's stated goal). Fixing it is a pure reliability/UX win with no new external dependencies for the single-instance case.

### Important scoping correction vs. the issue body

The issue text says "Client reconnects with `Last-Event-ID` header to resume" as if reconnect logic already exists — it doesn't. I read the current implementation and confirmed two things that change the shape of the client work:

[...]

PR #2047, opened 2026-07-14
Resumable chat SSE streams — buffer + Last-Event-ID replay

manzke:
## Summary

Implements #1497: SSE chat streams currently die completely on any network hiccup, with no reconnect logic at all, so a dropped connection mid-response loses the partial answer. This adds server-side event buffering with a monotonic per-chat event id and client-side reconnect-with-replay, so a brief network blip (page refresh, tab wake, WiFi drop) no longer silently loses an in-flight AI response.

- **Server**: every chat SSE event (`connected`, `chunk`, `thinking`, `citation`, `tool-call-*`, `done`, `error`, etc.) is now buffered per-chat with a monotonic id, in a new `MemoryStreamBuffer`/`StreamBufferService` (`server/services/streaming/`) — a sliding 5-minute TTL keyed off the last event, so an actively-streaming chat never expires mid-flight. `sendSSE` now writes an `id:` line when buffering is enabled.
- On GET reconnect (`server/routes/chat/sessionRoutes.js`), a `Last-Event-ID` header triggers a replay of whatever was buffered after that id, before the connection resumes live delivery.
- **Client**: `parseSseStream.js` now captures the `id:` field (previously intentionally ignored) and passes it to the caller. `useEventSource.js` tracks the last event id and auto-reconnects up to 3 times (500ms/1.5s/3s backoff) with `Last-Event-ID` on a dropped stream, before falling back to the existing user-facing error. A genuinely new message send (a fresh `initEventSource` call) is unaffected — it still resets state and notifies the server to stop any prior stream, exactly as before.
- Gated behind a new `resumableStreams` feature flag (`server/featureRegistry.js`, category `ai`, default **on**). No config migration needed — it's a registry default, not a persisted field, so nothing to seed for existing installs.

### Scoping decisions (matching the issue's own recommendations)

- **Workflow-driven chats are explicitly out of scope for the generic replay** — they already have a bespoke reconnect mechanism (`drainPendingFinish`/`buildReplayStepsFromState`), so the new buffer-replay is skipped whenever `activeWorkflowExecutions.has(chatId)` or a pending finish is stashed, to avoid double-delivering the same workflow step/result. `chatBridge.js` gained a small non-destructive `hasPendingFinish()` check for this.
- **Multi-instance/Postgres persistence** (mentioned in the issue) is descoped — this is in-memory, single-instance only, matching the issue's own suggestion to ship that as a follow-up once a shared-store layer exists.
- **A pre-flight "API key missing" error path** (`server/utils/ApiKeyVerifier.js`) writes directly to the response and bypasses `actionTracker`/buffering entirely — a narrow, already-existing gap (not introduced by this change) left as a follow-up rather than expanding scope.
- **No user-visible "connection was interrupted" badge** — descoped per the issue's own "nice to have" framing; the client silently resumes.

## Test plan

- [x] New `client/src/shared/utils/parseSseStream.test.js` — `id:` field capture, an event with no `id:` doesn't inherit the previous one's id, a read failure rethrows without parseSseStream emitting its own `error` event (so the caller can decide to retry silently), AbortError is swallowed. `node client/src/shared/utils/parseSseStream.test.js` (wired into `test:quick` as `test:parse-sse-stream`).
- [x] New `server/tests/streamBufferService.test.js` — monotonic per-chat ids, `eventsSince` ordering/filtering, unknown-chat returns `null` (vs. an empty replay), `clear()`, and the sliding-TTL eviction behavior. `node server/tests/streamBufferService.test.js`.
- [x] `npm run lint:fix && npm run format:fix` — clean.
- [x] Manual end-to-end verification against a running dev server (local admin login, real chat app): confirmed `id:` lines are emitted and monotonic across `connected`/`session.start`/`error`; confirmed a GET reconnect with `Last-Event-ID: <n>` replays exactly the missed events (no duplicates, no gaps) and then resumes live; confirmed a fresh connect with no `Last-Event-ID` is unaffected; confirmed setting `resumableStreams: false` in `features.json` (with a server restart, since config is cached in-memory) fully restores the previous byte-identical behavior (no `id:` lines at all).
- [x] Full server boot smoke test — no errors in logs, `/api/health` OK.

https://claude.ai/code/session_019u5HS9M7uaKLCZpATL74RY

---
_Generated by [Claude Code](https://claude.ai/code/session_019u5HS9M7uaKLCZpATL74RY)_


manzke:
Heads-up: #2302 (specifically #2304) formalizes reconnect/replay on top of the run ledger — overlaps with this PR's buffer + Last-Event-ID approach. Worth coordinating before merging both.

manzke:
Thanks for the heads-up. Looked at #2302/#2304 — real overlap, but I don't think it's a conflict yet: #2304 depends on #2303 (storage abstraction, not started) and is part of a multi-step epic, so it's likely a while before the ledger-based reconnect/replay actually lands and touches `sessionRoutes.js`'s disconnect handling.

This PR is a small, self-contained, feature-flagged (`resumableStreams`) in-memory buffer — no persistence, no schema, no dependency on anything else. It doesn't touch the disconnect-abort behavior in `sessionRoutes.js:374-389` that #2304 plans to change; it only adds buffering + `Last-Event-ID` replay around it.

Options as I see them:
1. **Merge this now as a stopgap** — gives users the "don't lose a partial answer on a network blip" benefit today with zero dependencies. When #2304's ledger replay lands, it supersedes this: `MemoryStreamBuffer`/`StreamBufferService`, the `id:`/`Last-Event-ID` wiring, and the `resumableStreams` flag would be removed in that PR in favor of ledger-backed replay.
2. **Hold/close this PR now** and let #2304 deliver reconnect/replay as part of the epic, accepting the current lossy-on-disconnect behavior until then (likely weeks-to-months out given the dependency chain).
3. Something else you have in mind (e.g. land this but explicitly scope #2304 to build on top of / replace it rather than duplicate).

I lean toward (1) since it's low-risk and already implemented/tested, but this is your call given you're the one scoping #2304 — let me know which way to go and I'll act on it (or just close this if you'd rather wait for the ledger approach).

---

Issue #1781, opened 2026-07-06
Split the 2545-line AppFormEditor into section components with a shared nested-update helper


manzke:
**Severity:** medium · **Confidence:** suggestion · **Effort:** large
**Category:** architecture

### Problem
AppFormEditor is a single 2545-line component: one giant return with 15+ config sections (basic info, system, iframe/redirect, variables, starter prompts, websearch, iassistant, tools, MCP, workflows, skills, upload x3, input mode, greeting) and a dozen hand-written immutable-update handlers repeating the same map/spread pattern. PlatformFormEditor (1840 lines) repeats the same shape with its own add/update/remove triplets per provider type (OIDC/JWT/LDAP, lines 136-275) and mostly untranslated hardcoded strings. Any change requires scrolling a 2.5k-line file; the same handler bug must be fixed N times.

### Evidence
AppFormEditor.jsx:150-236 four near-identical nested-array handlers; :1239-1330 tools/MCP sections; :1629-1930 three parallel upload sub-forms. PlatformFormEditor.jsx:136-275 three add/update/remove provider triplets differing only in config key.

### Recommended fix
Extract each card into components/app-form/ section files receiving (app, onChange), plus a tiny updateIn(obj, path, value) helper (or use the existing DynamicLanguageEditor pattern) to kill the repeated spread pyramids. Do the same provider-list extraction for PlatformFormEditor.

### Affected files
- `client/src/features/admin/components/AppFormEditor.jsx:36`
- `client/src/features/admin/components/PlatformFormEditor.jsx:70`

---
<sub>From the 2026-07-06 comprehensive review · tracking #1677 · report `concepts/2026-07-06 Comprehensive Code Review.md`</sub>

manzke:
## Implementation Plan

**What:** `AppFormEditor.jsx` (2611 lines) and `PlatformFormEditor.jsx` (1840 lines) are each one giant component holding 15+ config sections and a dozen hand-written immutable-update handlers that repeat the same map/spread pattern — split each into per-section components sharing a generic nested-update helper, so a section can be found, tested, and fixed without scrolling a 2.5k-line file.

[...]

Issue #1740, opened 2026-07-06
Consolidate copy-paste CRUD across apps/models/prompts/tools/sources into a resource-route factory — drift is already causing bugs


manzke:
**Severity:** medium · **Confidence:** verified · **Effort:** large
**Category:** duplication

### Problem
apps.js, models.js, prompts.js, tools.js and sources.js each hand-roll the same list/get/create/update/delete/toggle/batch pipeline (validateIdForPath → configCache read → validate → write → refresh cache → snapshot → audit) and have drifted: apps PUT uses atomicWriteJSON while apps POST/toggle and all of models/prompts/tools use plain fs.writeFile (crash mid-write corrupts config); sources duplicates identical validateIdForPath blocks in three routes; models re-implements 'exactly one default model' five ways despite modelsLoader.ensureOneDefaultModel; Zod schemas exist but writes never enforce them; responseHelpers' asyncHandler/createRouteHandler are exported but unused. resourceLoader.js proves the factory pattern already worked for loading.

### Evidence
apps.js:623 `await atomicWriteJSON(appFilePath, updatedApp)` vs models.js:203 `await fs.writeFile(modelFilePath, JSON.stringify(updatedModel, null, 2))` and prompts.js:546 same; sources.js:558-566 two consecutive identical `validateIdForPath(id, 'source', res)` blocks (also 836-843, 988-995); models re-implements default-model logic at 186-196, 270-280, 312-325, 376-382, 416-428; grep shows zero usages of asyncHandler/createRouteHandler outside responseHelpers.js; resourceLoader.js:13 "This eliminates the duplication between appsLoader, modelsLoader, and promptsLoader."

### Recommended fix
Build a createAdminResourceRoutes({resource, dir, schema, hooks}) factory (mirror of createResourceLoader) that owns validation (including Zod enforcement on write), atomic writes, cache refresh, snapshot/audit, and route ordering; migrate apps/models/prompts first, then tools/sources with per-resource hooks.

### Affected files
- `server/routes/admin/apps.js:623`
- `server/routes/admin/models.js:203`
- `server/routes/admin/prompts.js:546`
- `server/routes/admin/tools.js:438`
- `server/routes/admin/sources.js:558`
- `server/utils/responseHelpers.js:103`

---
<sub>From the 2026-07-06 comprehensive review · tracking #1677 · report `concepts/2026-07-06 Comprehensive Code Review.md`</sub>

manzke:
# Implementation Plan

## What
[...]

Issue #1705, opened 2026-07-06
Wire the unit/integration test suites into CI — nothing runs today


manzke:
**Severity:** high · **Confidence:** suggestion · **Effort:** medium
**Category:** testing

### Problem
The repo contains ~134 files in server/tests, plus tests/unit, tests/integration and tests/e2e, but no GitHub workflow ever executes them. The only test run in CI is the accessibility Playwright suite; every other workflow builds, lints, or packages. With zero enforcement, the suites have silently rotted: tests reference deleted paths and incompatible runners without anyone noticing. An unenforced ~20k-line test corpus is a liability, not an asset.

### Evidence
grep for 'npm run test' across .github/workflows/*.yml matches only accessibility.yml:47 ('run: npm run test:a11y -- --project=chromium'). build-binaries.yml, docker-ci.yml, auto-lint-format.yml contain no test step. package.json:76-88 defines test:all/test:unit/test:integration and server/package.json:10 defines a jest run — none referenced in CI. server/tests/*.test.js totals 19,575 lines.

### Recommended fix
Add a PR workflow that runs the subset of tests that actually pass (start with tests/unit and the node:test files via `node --test`), then ratchet up. Delete or quarantine suites that cannot be made green so the signal stays trustworthy.

### Affected files
- `.github/workflows/accessibility.yml:47`
- `package.json:76`
- `server/jest.config.js:14`

---
<sub>From the 2026-07-06 comprehensive review · tracking #1677 · report `concepts/2026-07-06 Comprehensive Code Review.md`</sub>

manzke:
## Implementation Plan

**What:** CI still doesn't run most of the project's test corpus — only a slice does. `unit-tests.yml` was added on 2026-07-07 (after this issue was filed) and now runs `test:quick` (`test:unit` + `test:adapters` + `test:token-stats` + `test:export-formats` + `test:feature-flags`) on every PR, but `test:integration`, `test:e2e` (beyond the single `accessibility.spec.js`), and the entire 154-file `server/tests/` Jest corpus (its own config/runner, `server/jest.config.js` + `server/package.json`'s `test` script) still have zero CI enforcement.

**Why:** An unenforced test suite is worse than no suite — it creates false confidence ("we have tests") while regressions ship silently. `server/tests/` alone is ~150 files that could bit-rot indefinitely without anyone noticing (imports breaking, assertions going stale), and the same already happened to two files under `tests/integration/` (see below).

[...]

Issue #1499, opened 2026-05-20
[Platform] Multi-instance horizontal scaling via PostgreSQL LISTEN/NOTIFY


manzke:
## Phase 3B — Multi-Instance Scaling

Enable running multiple ihub-apps instances behind a load balancer. All coordination via PostgreSQL — no Redis needed.

### Distributed ConfigCache Invalidation
- PostgreSQL `LISTEN/NOTIFY` channel `ihub_config_invalidate`
- Admin saves config → NOTIFY → all instances invalidate cache
- Single-instance fallback: TTL-based refresh continues to work

### SSE Notification Fan-Out
- PostgreSQL `LISTEN/NOTIFY` channel `ihub_notify_{userId}`
- Any instance can notify any user; the instance holding the SSE connection delivers
- No sticky sessions needed at load balancer level

### Multi-Instance Auth Stores
- Move in-memory OAuth stores (`authorizationCodeStore.js`, `consentStore.js`, `refreshTokenStore.js`) to PostgreSQL tables when DATABASE_URL is set
- Continue as in-memory Maps for single-instance deployments

### Docker Compose
```yaml
services:
  ihub-app:
    image: ihub-apps:latest
    command: node server/server.js
    deploy: { replicas: 3 }
    environment:
      - DATABASE_URL=postgres://...
  ihub-worker:
    image: ihub-apps:latest
    command: node server/worker.js
    deploy: { replicas: 5 }
    environment:
      - DATABASE_URL=postgres://...
      - WORKER_CONCURRENCY=4
  postgres:
    image: postgres:16
```

**Dependencies:** Requires #1490 (PostgreSQL), #1495 (job queue/workers).

manzke:
## Implementation Plan

**What:** Enable running multiple ihub-apps instances behind a load balancer by coordinating config-cache invalidation, SSE notification delivery, and OAuth flow state entirely through PostgreSQL `LISTEN`/`NOTIFY` — no Redis or other coordination service required.

[...]

PR #2053, opened 2026-07-14
feat(persistence): add StorageProvider abstraction (filesystem + Postgres)

manzke:
## Summary

Phase 1A foundation for #1490 ([Platform] PostgreSQL + persistence abstraction layer): introduces a `StorageProvider` abstraction so configuration reads go through one interface instead of scattered `fs` calls, with a `FilesystemProvider` (default, byte-identical to today's behavior) and an opt-in `PostgresProvider` (activated via `DATABASE_URL`). This is the prerequisite several other open issues are blocked on (#1499 multi-instance scaling, part of #1497's Postgres fallback).

## What changed

- **`server/persistence/`** — `StorageProvider.js` (interface), `FilesystemProvider.js` (wraps `contents/` on disk, same resolution/traversal-guard logic `configLoader.js` used before), `PostgresProvider.js` (stores each path as a row in a generic `config_kv` table), `StorageRegistry.js` (factory: Postgres when `DATABASE_URL` is set, filesystem otherwise, memoized per process).
- **`server/db/`** — `pool.js` (lazy `pg.Pool` singleton, `null` when `DATABASE_URL` is unset — dynamically imports `pg` so filesystem-only deployments are unaffected), `schema.js` (idempotent `config_kv` DDL, applied lazily on first Postgres use).
- **`server/configLoader.js`** — `loadFile` now reads through `getStorageProvider()` instead of `fs.readFile` directly. `loadBuiltinLocaleJson`/`listBuiltinLocales` (which read from `shared/i18n/`, not `contents/`) are untouched.
- **`server/config.js`** — new optional `DATABASE_URL` env var.
- **New migration `V078__add_database_settings.js`** — seeds `platform.json`'s `database.enabled: false` default for existing installs; `server/defaults/config/platform.json` ships it directly for fresh installs.
- **`docs/persistence.md`** (new, linked from `docs/SUMMARY.md`) documents the current scope and how to opt in.
- Added `pg` as a server dependency.

## Deliberately out of scope (follow-up)

Per the open questions already raised in #1490's implementation-plan comment, this PR does **not** attempt:
- Migrating admin write paths (`server/routes/admin/*.js`) off `atomicWriteJSON` onto `StorageProvider.write()` — that's a much larger, separate-decision change (schema-enforcement-on-write, response-shape questions already raised on related issues #1740/#1703).
- The "Database" admin UI page / test-connection endpoint.
- A `contents/` → PostgreSQL data-migration tool.
- Docker Compose wiring for a `postgres` service.

None of those are needed for the filesystem-default path every existing deployment uses today, and bundling them here would make for an unreviewable diff — same split this repo's convention already uses for other large issues (e.g. #1781, #1740).

## Test plan

- [x] New unit tests: `FilesystemProvider` (tmp-dir backed), `PostgresProvider` (fake in-memory pool, no real DB needed), `StorageRegistry` (provider selection + memoization), `db/pool.js` (null without `DATABASE_URL`, memoized `pg.Pool` construction) — all passing.
- [x] Ran the full server test suite before/after this change: identical baseline (114 failing suites / 45 failing tests, pre-existing and unrelated — see #1705/#1706) — this PR introduces **zero new test failures**.
- [x] `npx eslint` / `npx prettier --check` clean on all changed/new files.
- [x] Manually booted the dev server (4 workers) with no `DATABASE_URL` set: boots cleanly, migration `V078` applies successfully, `platform.json` gets `database.enabled: false`, and a `loadJson()`/missing-file smoke test confirms `configLoader.js` behavior is unchanged.

https://claude.ai/code/session_01XsoPForsrFp9Gi94b7HSoe

---
_Generated by [Claude Code](https://claude.ai/code/session_01XsoPForsrFp9Gi94b7HSoe)_

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -19,6 +19,7 @@
     - [Mimetype Configuration](mimetypes.md)
     - [Configuration Validation](configuration-validation.md)
     - [Configuration Migrations](configuration-migrations.md)
+    - [Persistence Layer](persistence.md)
     - [Localization](localization.md)
   - [Authentication & Security]()
     - [Authentication Architecture](authentication-architecture.md)
--- a/docs/persistence.md
+++ b/docs/persistence.md
@@ -1,0 +1,76 @@
+# Persistence Layer
+
+iHub Apps stores its configuration (`contents/config/*.json`, apps, models,
+prompts, tools, sources, ...) behind a small `StorageProvider` abstraction
+rather than calling `fs` directly everywhere. This is the foundation for
+running iHub against PostgreSQL instead of the local filesystem, which in
+turn is a prerequisite for multi-instance horizontal scaling (distributed
+config-cache invalidation, shared OAuth state, etc.).
+
+## Current scope
+
+This is a **foundation layer**, not a full feature yet:
+
+- **Filesystem is still the default and only fully-wired backend.** Every
+  existing deployment continues to read/write `contents/` exactly as before
+  — nothing changes unless `DATABASE_URL` is set.
+- **Reads** for cached JSON/text configuration (`configLoader.js` — apps,
+  models, prompts, groups, platform config, etc.) already go through the
+  active `StorageProvider`.
+- **Writes** from the admin UI (`server/routes/admin/*.js`) still use
+  `atomicWriteJSON` directly against the filesystem. Migrating every admin
+  write path onto `StorageProvider` — and building the "Activate PostgreSQL"
+  admin UI, data-migration tool, and Docker Compose service — is tracked as
+  follow-up work.
+
+## Architecture
+
+```
+server/persistence/
+  StorageProvider.js     — interface: read, write, delete, list, exists
+  FilesystemProvider.js  — wraps contents/ on disk (default, backward-compatible)
+  PostgresProvider.js    — stores each path as a row in a generic config_kv table
+  StorageRegistry.js     — factory: PostgresProvider when DATABASE_URL is set, else FilesystemProvider
+
+server/db/
+  pool.js     — lazy pg.Pool singleton; null when DATABASE_URL is unset
+  schema.js   — DDL for config_kv (idempotent, applied on first use)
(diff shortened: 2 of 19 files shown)


manzke:
Fixed the CodeQL path-injection alert on `FilesystemProvider.js` in 8f6221c: added an explicit, synchronous containment re-check (`path.resolve` + `startsWith`) on the resolved path immediately before it's used in any `fs.*` call. `resolveAndValidatePath()` (in `pathSecurity.js`) already sanitizes correctly, but its guarantee is behind an async `fs.realpath`-based call in another module that static analysis can't trace through — this makes the check self-contained and statically verifiable in this file. Added a regression test covering several nested traversal attempts.

---

manzke:
The previous fix didn't clear the CodeQL alert because it addressed the wrong spot. Root cause found in d264404: on a blocked traversal attempt, `_resolve()` built a fallback path via `path.join(baseDir, path.basename(relativePath))` — `path.basename()` is a taint-preserving operation over the untrusted input, so this reconstructed a fresh tainted path expression regardless of any containment check performed afterward on the *other* (valid) branch.

Fix: removed that fallback entirely. `_resolve()` now fails closed (throws, tagged `error.code = 'ENOENT'`) on a blocked path instead of deriving a substitute from untrusted input:
- `read()`/`exists()`/`list()` catch it → `null`/`false`/`[]` (same "not found" contract as before, and in practice what the old clamped path almost always resolved to anyway).
- `write()`/`delete()` let it propagate — a mutation silently no-op'ing or landing somewhere unexpected is the wrong failure mode.

Updated tests accordingly; verified `configLoader.js` behavior is still unchanged for legitimate reads via a manual smoke test.

---

manzke:
Closing: 113+ commits behind main and the design has moved past the config_kv model — see #2302 (design doc in concepts/persistence-layer/). The filesystem path-containment hardening iterated on this branch is carried into #2303's requirements so that work isn't lost.

PR #2149, opened 2026-07-28
fix(oauth): resolve authorization codes across cluster workers

manzke:
## Problem

OAuth login for MCP clients fails right after the consent screen with `invalid_grant: Authorization code is invalid or expired`.

Since round-robin routing landed (#2146), the two halves of the authorization code flow run on different workers. `POST /api/oauth/authorize/decision` mints the code into a per-process `Map` on worker A; the client's `POST /api/oauth/token` is handed to the next worker, whose map is empty.

**This is not intermittent.** Round-robin gives each new connection to the next worker in sequence, so two back-to-back requests land on *adjacent* workers and the exchange fails ~100% of the time, not `(N-1)/N`. Measured against a real 4-worker cluster:

| | token exchanges succeeded |
| --- | --- |
| before | **0 / 12** — all `invalid_grant: Authorization code is invalid or expired` |
| after | **12 / 12** |

`npm run dev` pins `WORKERS=1`, which is why this never showed up in development.

The same flow had a second instance of the same bug: the consent CSRF token and the PKCE `code_challenge` lived in an `express-session` backed by a per-worker `MemoryStore`, so a decision POST reaching another worker saw `403 CSRF token missing` — or, when CSRF happened to pass, minted a code with an **empty** `codeChallenge` that the token endpoint then rejected.

## Approach

**Codes stay on the minting worker; the consume is routed to it.** Replicating codes to every worker would fix the lookup and destroy the security property — reading a code *consumes* it, so N copies means N valid redemptions. Ownership is announced over the existing cluster bus and a worker receiving the token request asks the owner to consume it. Exactly one process ever holds a code, so single-use stays atomic with no distributed agreement.

This needed a request/reply primitive on the bus (`request()` / `respond()`), built on the existing pub/sub with correlation ids so the primary stays a dumb repeater. A handler returning `undefined` sends no reply, which is what makes the owner-unknown broadcast fallback safe: non-owners stay silent instead of racing a `null` ahead of the real owner's answer.

`consumeCode()` is now async; the single call site awaits it.

**A code is now a routing handle and a secret, dot-separated** (`handle.secret`) — 128 bits of handle, 256 bits of secret. Ownership announcements are broadcast to *every* worker and retained for the code's lifetime, so they carry only the handle, which grants nothing on its own. The secret crosses the IPC boundary once, in the single directed consume message to the owning worker, which verifies it with `timingSafeEqual` and destroys it — unavoidable in any design, since only the owner can verify, and a different exposure from a retained broadcast. A valid handle presented with a wrong secret discards the entry, so a guessing loop gets one attempt per code rather than many. Codes are opaque to clients per RFC 6749, so the shape is not a contract.

**Consent context moves into a signed ticket, not a session.** The whole context travels in the form protected by an HMAC, so any worker can verify it with no shared state. It replaces the CSRF token by being bound to the user it was issued for — a ticket cannot be forged without the signing key, and one obtained by an attacker cannot drive a victim's consent. Signing fails closed: with no JWT secret configured it refuses rather than HMAC-ing with an empty key, and verification rejects cleanly instead of raising a 500.

Side benefit: `redirect_uri`, `scope`, `code_challenge` and `nonce` are now integrity-protected instead of trusted from the POST body, so a client cannot widen scope or redirect the code between the screen and the decision.

## Tests

`server/tests/authCodeCluster.test.js` forks a real 3-worker cluster and drives the two halves of the exchange on *different* workers — mocking `node:cluster` would only test the mock. Covers cross-worker consume, replay rejection from a third worker, rejection of a forged secret by the owning worker, ownership retraction once a code is spent, and that announcements carry only the routing handle and never the full code.

`server/tests/consentTicket.test.js` covers round-trip, tamper detection on payload and signature, forging with a wrong key, expiry, malformed input, and the no-signing-key path. It initialises the secret through `TokenStorageService` the way `server.js` does, so it exercises the real resolution chain.

Both are wired into `npm run test:cluster`. Also verified end-to-end against a live 4-worker cluster: 10/10 full exchanges across both the consent-screen and remembered-consent paths, with code replay rejected every time.

## Also in this PR

- Corrects `docs/mcp-integration.md`, which told readers stateless mode was only needed for multi-replica deployments and that the sticky router keeps a client on one worker. Neither holds since #2146 — the MCP gateway's own session map has this same bug, so stateless mode is needed on any deployment running more than one worker.
- `concepts/2026-07-28 Round-Robin Worker-Local State Audit.md` records the **remaining** instances of this bug class found while investigating, with severity and recommended sequencing. **None of them are fixed here** — flagging them for follow-up:

  | Area | Symptom |
  | --- | --- |
  | MCP gateway sessions | `404 Session not found` on the first tool call after `initialize` |
  | OIDC session store | SSO login fails `Failed to verify request state` (PKCE verifier + state lost) |
  | Integration OAuth sessions | Jira / Office 365 / Google Drive / Nextcloud connect fails `invalid_state` |
  | Short links | Created link 404s on 3 of 4 workers *permanently*; writes clobber other workers' links |
  | Workflow execution registry | `404 Execution` on stream attach / cancel / checkpoint reply |
  | OCR job store | Progress stuck at 0%, download 404s |
  | Rate limiters | Per-worker counters, so the effective limit is the configured value times the worker count (≈200 login attempts/15min instead of 50) |
  | iAssistant conversation state | Context lost mid-chat; a new upstream conversation per turn |

## Notes for the reviewer

- **CodeQL raised `js/insufficient-password-hash` on an earlier revision** that keyed the store on a SHA-256 of the code. Inline suppression is not available — GitHub code scanning has no inline suppression for CodeQL, and `lgtm[...]` was LGTM.com syntax, which is retired, so the existing `lgtm[...]` comments in `pkceUtils.js` and `refreshTokenStore.js` are decorative. The handle/secret split removes the sink rather than suppressing a false positive, and is the better design for the reasons above. Both alerts are now resolved.
- **`config.JWT_SECRET` does not exist.** `JWT_SECRET` is not declared in `server/config.js`, so priority 1 of `resolveJwtSecret()` is dead code, as is the same branch at `middleware/setup.js:260` for the session secret. The value actually reaches the platform via `TokenStorageService.initializeJwtSecret()`, which reads `process.env.JWT_SECRET` directly. Not changed here — worth a follow-up.
- **`mcpServer.requireConsent` does not force per-grant consent.** `forceConsent` bypasses the trusted-client shortcut but remembered consent still short-circuits below it, so the stated intent ("explicit per-grant consent is warranted") is not achieved. Left alone because fixing it changes behaviour.
- With `WORKERS=4` all four workers race the migration lock at boot, so three log `Configuration migration failed` at error level. Harmless — one worker applies migrations — but misleading in logs. Pre-existing.

## Verification

- `npm run test:cluster` — all 6 suites pass
- `npx eslint server/` — 0 errors
- `npm run test:unit` — 245 tests pass (6 client suites fail on missing `client/node_modules` in the test container; no client code touched)
- `tests/manual/oauth-auth-code-unit.js` — 35 pass, 2 pre-existing `jwtAuth.js` structural failures unrelated to this change (confirmed present on the base commit)
- Server boots clean with `WORKERS=4`

### DIFF
--- a/concepts/2026-07-28 Round-Robin Worker-Local State Audit.md
+++ b/concepts/2026-07-28 Round-Robin Worker-Local State Audit.md
@@ -37,8 +37,16 @@ Replicating codes to all workers would have destroyed the security property — 
 *consumes* it, so N copies means N valid redemptions. Instead the code stays on the minting worker,
 ownership is announced over `server/clusterBus.js`, and a worker receiving the token request asks
 the owner to consume it. Exactly one process ever holds a code, so single-use stays atomic with no
-distributed agreement. Only the SHA-256 of the code is announced, so the raw credential never
-leaves the worker that minted it.
+distributed agreement.
+
+A code is now `<handle>.<secret>` — 128 bits of routing handle, 256 bits of secret. Ownership
+announcements are broadcast to every worker and retained for the code's lifetime, so they carry only
+the handle, which grants nothing on its own. The secret crosses the IPC boundary once, in the single
+directed consume message to the owner, which verifies it in constant time and destroys it. An
+earlier revision announced a SHA-256 of the whole code instead; the split is better because the
+broadcast value was never secret to begin with, and it avoids hashing a request-derived credential
+(which CodeQL's `js/insufficient-password-hash` flags, and GitHub code scanning cannot suppress
+inline — the `lgtm[...]` comments elsewhere in this repo are decorative, LGTM.com is retired).
 
 This required a request/reply primitive on the bus (`request()` / `respond()`), built on the
 existing pub/sub with correlation ids so the primary remains a dumb repeater.
--- a/concepts/2026-07-28 Round-Robin Worker-Local State Audit.md
+++ b/concepts/2026-07-28 Round-Robin Worker-Local State Audit.md
@@ -1,0 +1,107 @@
+# Round-Robin Worker-Local State Audit
+
+**Date:** 2026-07-28
+**Trigger:** OAuth login for MCP clients failed with `invalid_grant: Authorization code is invalid or expired`
+**Status:** OAuth authorization code flow fixed; remaining findings open
+
+## Root cause class
+
+`server/server.js` forks `config.WORKERS` workers (default **4**). Since #2146 the default
+connection routing is **round-robin** (`STICKY_SESSIONS` defaults to `false`), so consecutive
+requests from the same client are handled by *different* worker processes.
+
+Any mutable state kept in a worker's own process memory is therefore invisible to the requests that
+follow. Where request A writes state and a later request B reads it, B fails.
+
+Two properties make this worse than it first looks:
+
+- **It is not "1 in N".** Round-robin hands each new TCP connection to the next worker in sequence.
+  Two requests made back-to-back over separate connections land on *adjacent* workers, so a
+  two-request handshake fails close to **100%** of the time rather than `(N-1)/N`. This was measured:
+  the OAuth token exchange failed 12/12 against a 4-worker cluster before the fix.
+- **`npm run dev` pins `WORKERS=1`.** None of this reproduces in development. It appears only in
+  production, `start:prod`, Docker, and binary deployments.
+
+`#2148` (config cache) and `#2146` (chat/SSE state) fixed two instances of this class. The findings
+below are the remainder.
+
+## Fixed in this change
+
+### OAuth authorization codes — `server/utils/authorizationCodeStore.js`
+
+`POST /api/oauth/authorize/decision` minted the code into a per-process `Map`; the client's
+`POST /api/oauth/token` arrived at another worker whose map was empty, and
+`server/routes/oauth.js` returned `invalid_grant: Authorization code is invalid or expired`.
+
+Replicating codes to all workers would have destroyed the security property — reading a code
+*consumes* it, so N copies means N valid redemptions. Instead the code stays on the minting worker,
+ownership is announced over `server/clusterBus.js`, and a worker receiving the token request asks
+the owner to consume it. Exactly one process ever holds a code, so single-use stays atomic with no
+distributed agreement. Only the SHA-256 of the code is announced, so the raw credential never
+leaves the worker that minted it.
+
+This required a request/reply primitive on the bus (`request()` / `respond()`), built on the
+existing pub/sub with correlation ids so the primary remains a dumb repeater.
+
+### Consent screen CSRF + PKCE — `server/routes/oauthAuthorize.js`
+
+The consent CSRF token and the PKCE `code_challenge` lived in an `express-session` backed by a
+per-worker `MemoryStore` (`server/middleware/setup.js`). A decision POST reaching another worker
+saw no session, giving `403 CSRF token missing` — or, if CSRF happened to pass, a code minted with
+an **empty** `codeChallenge` that the token endpoint later rejected.
+
+Replaced with a signed, self-contained consent ticket (`server/utils/consentTicket.js`) carried in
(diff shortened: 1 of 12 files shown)


manzke:
Fixed in 01fb217 — the CodeQL alert was real in the sense that my suppression comment didn't apply: I put `lgtm[js/insufficient-password-hash]` on the line *above* the hash call, and suppression only works on the flagged line itself. Moved to the end of the line with a justification, matching the identical case at `server/utils/refreshTokenStore.js:87`.

The alert itself is a false positive. CodeQL traces the taint from the `code` request parameter and treats it as a password, but:

- The value is a server-minted `crypto.randomBytes(32)` authorization code — 256 bits of CSPRNG output, never user-chosen — with a 10-minute TTL and single use.
- The digest is an index key for the store/presence map, not a stored credential verifier.
- Key stretching (bcrypt/scrypt/argon2) exists to make offline brute force of *low-entropy* secrets expensive. Against 256 random bits it buys nothing; guessing the code is already infeasible.
- Hashing was introduced here specifically as a hardening measure: it means the raw code is never sent over the cluster IPC bus and never appears in another worker's memory. Removing the hash to satisfy the query would make the code strictly less safe.

---

manzke:
Thanks — three of the four were right and are fixed in 1894660.

**`consentTicket.js:66` — empty HMAC key fallback.** Correct and the most serious of the batch. `String(resolveJwtSecret() || '')` meant a server with no JWT secret would sign with `''`, which anyone can reproduce — forgeable tickets. Now throws.

**`consentTicket.js:112` — verify could throw.** Correct, and it became reachable once signing throws. `verifyConsentTicket()` now catches, logs it as the server misconfiguration it is, and rejects cleanly rather than turning it into a 500 mid-consent.

While fixing this: `config.JWT_SECRET` doesn't exist — `JWT_SECRET` is not declared in `server/config.js`, so priority 1 of `resolveJwtSecret()` is dead code. The value actually arrives via `TokenStorageService.initializeJwtSecret()`, which reads `process.env.JWT_SECRET` directly. `middleware/setup.js:260` has the same dead branch for the session secret. Left alone — out of scope here, but worth a follow-up.

**`authorizationCodeStore.js:110/127/131` — dead `used` flag.** Correct, and your low-confidence comment identified it precisely: entries are deleted on first consume, so `used` was never `true` and the replay branch was unreachable. It was carried over from the original implementation. Removed both the field and the branch. Worth noting the replay *protection* was never affected — a replayed code is rejected because it is gone from the store, which is logged; only the more specific warning was dead.

**The CodeQL alert** (separate bot, same file) needed a different resolution, and it changed the design here. Inline suppression does not work: GitHub code scanning has no inline suppression for CodeQL, and `lgtm[...]` was LGTM.com syntax, which is retired — so the existing `lgtm[...]` comments in this repo (`pkceUtils.js`, `refreshTokenStore.js`) are decorative. The alert stayed red after I moved mine onto the flagged line.

So instead of suppressing a false positive, the sink is gone. A code is now `<handle>.<secret>` — 128-bit routing handle, 256-bit secret — and the store keys on the handle. This is better than the digest it replaces, not just quieter:

- Ownership announcements are **broadcast to every worker and retained** for the code's lifetime. They now carry only the handle, which grants nothing alone. Previously they carried a digest of a live credential.
- The secret crosses IPC once, in the single directed consume message to the owning worker, which verifies with `timingSafeEqual` and destroys it. Unavoidable in any design — only the owner can verify — and a very different exposure from a retained broadcast.
- A valid handle with a wrong secret now discards the entry, so a guessing loop gets one attempt per code instead of many.

Codes are opaque to clients per RFC 6749, so the shape isn't a contract.

Re-verified end-to-end on a live 4-worker cluster after all of the above: 10/10 full exchanges across both consent paths, 12/12 token exchanges, replay rejected every time.

---

Issue #2217, opened 2026-08-24
`server/tests/*.test.js` never run in CI — jest testMatch does not cover them


manzke:
Split out of #2213 (Q9 — "create a github issue for it").

## What is wrong

`tests/config/jest.config.js` matches only three locations:

```js
testMatch: [
  '**/tests/integration/**/*.test.js',
  '**/tests/unit/server/**/*.test.js',
  '**/tests/unit/client/**/*.test.jsx'
]
```

Everything under `server/tests/` is therefore invisible to `npm run test:unit`, `npm run test:integration` and `npm run test:coverage`. `collectCoverageFrom` even excludes it explicitly (`'!server/tests/**'`).

Some of those files are wired into `package.json` individually as plain `node` scripts (`test:adapters`, `test:cluster`, `test:openai`, …), but the ones that are not run by any script are simply dead. `server/tests/auditLogService.test.js` is the case that surfaced in #2213: it exists, it is a jest-style suite, and it has never executed in CI.

## Why it was not fixed in #2213

Enabling `server/tests/**` in `testMatch` is likely to surface pre-existing failures across unrelated subsystems, which would have derailed an already sizeable PR. New tests written for #2213 were placed under `tests/unit/**` instead, so they actually run.

## What needs deciding

1. **Inventory.** Which files under `server/tests/` are jest suites, which are standalone `node` scripts, and which are stale? Some are integration tests that hit real LLM providers and must not run in CI regardless (`real-llm-integration.test.js`, `azure-openai-test.js`).
2. **For the jest-style suites:** move them under `tests/unit/server/` (consistent with the rest) or widen `testMatch` to include `server/tests/**`. Moving them is probably cleaner — one home for jest suites, and `collectCoverageFrom` stops needing an exclusion.
3. **Fix or delete what then fails.** Expect some breakage; several of these predate refactors.
4. **For the standalone `node` scripts:** either keep them as explicit `npm run test:*` entries (already the case for the adapter and cluster ones) or convert them to jest.
5. **Watch for `import.meta`.** babel-jest's CJS transform cannot evaluate it, so any suite whose import chain reaches `server/pathUtils.js` or `server/utils/authorization.js` needs those modules mocked — see `tests/unit/server/auditLogQuery.test.js` for the pattern.

No user-visible behaviour is affected; this is test infrastructure.

Issue #2216, opened 2026-08-24
Audit log: the `mcp` source is allowed but never written


manzke:
Split out of #2213 (Q7 — "remove for now, create a new GitHub issue for it with details").

## What is wrong

`server/validators/auditEntrySchema.js` declares four audit sources:

```js
export const auditSources = ['web', 'mcp', 'api', 'admin'];
```

but **nothing ever writes `mcp`**:

- `deriveSource(req)` in `server/services/AuditLogService.js` only ever returns `api` (OAuth client / `authMethod === 'oauth'`), `admin` (URL contains `/api/admin/`), or `web` (everything else).
- No call site passes `source: 'mcp'` explicitly.

So requests arriving through the MCP gateway are recorded as `web` or `api`, indistinguishable from browser and REST traffic. An admin cannot answer "what did MCP clients do?" from the audit log at all.

## What #2213 did about it

Nothing was removed from the schema — `auditSources` still allows `mcp`, so wiring it up later needs no schema change or migration. What changed is that the audit log's **source filter is now facet-driven**: its options come from the values actually present in the log, so `mcp` no longer appears as a pickable option that can only ever return zero rows.

## What is left to do

Decide what MCP auditing should record, then wire it up:

1. **Set the source.** Either extend `deriveSource()` to recognise the MCP request path, or pass `source: 'mcp'` explicitly from the MCP request handlers. Preferring an explicit pass is probably safer — `deriveSource`'s current URL sniffing is order-sensitive (`/api/admin/` wins over everything).
2. **Decide what is auditable.** Tool invocations, app/workflow calls, and the OAuth token exchanges are the obvious candidates. Today MCP tool calls are not audited at all, so this is more than a label change.
3. **Actor identity.** MCP callers are OAuth clients acting on behalf of a user. The entry should make clear which client and which user — `actor` currently carries one identity.
4. **Add the German/English value label** for the new source once it can occur (`admin.auditLog.values.source.mcp` in `shared/i18n/{en,de}.json`).

No admin action or config change is needed either way; this is purely about the audit log gaining MCP coverage.

PR #2218, opened 2026-08-24
fix(admin): audit log filters now apply, with checkbox multi-select (#2213)

### LINKED ISSUE #2213: Audit Log Filtering not possible
## Original report

> The audit log filtering is not working. If you select a resource, a method or anything else the log is not filtered. Also it can be cumbersome to filter, because often you have a lot of logins, but you only want to see the others.
> We should improve the filters to be a checkbox based list where I can deselect certain value and/or none/all at once.

---

## Root cause — why no filter does anything

`client/src/features/admin/pages/AdminAuditLogPage.jsx:328` calls two `useFilterState` setters in the same tick:

```js
const handleFilterChange = setter => value => {
  setter(value);        // setSearchParams(fn) -> navigate("?resource=app")
  setPageParam('1');    // setSearchParams(fn) -> navigate("?")   <-- overwrites the line above
};
```

React Router documents that the function form of `setSearchParams` **does not queue**:

> The function callback version of `setSearchParams` does not support the queueing logic that React's `setState` implements. Multiple calls to `setSearchParams` in the same tick will not build on the prior value.

The v7 source confirms it — the updater receives a copy of the **render-time** `searchParams` closure, not a live ref. So `setPageParam('1')` recomputes from the params as they were *before* the filter was applied, deletes `page` (it equals its default), and navigates. The filter never reaches the URL, the `<select>` snaps back to "All", and `fetchEntries` never re-runs.

**Only this page is affected.** Every other `useFilterState` consumer (`AdminUsersPage`, `AdminAppsPage`, `AdminMarketplacePage`, …) sets exactly one parameter per change. This page is the only one chaining two.

**Same bug, second location** — `AdminAuditLogPage.jsx:526`: changing the page size silently reverts to 50.

**The server is fine.** `queryAuditLog()` filters correctly on all five fields; the parameters simply never arrive.

## Additional defects found

1. **The resource dropdown is provably wrong.** It offers `provider` — which nothing writes — and omits `tool`, `credential`, `integrations`, `marketplaceItem`, `marketplaceRegistry`, `uiConfig`, `uiAsset`, `toolScript`, `apps`, plus every path-derived resource the audit middleware produces (`usage`, `configs`, `workflows`, …). `auditEntrySchema.js` documents `resource` as deliberately free-form, so a hardcoded list can never be correct.
2. **The actor dropdown only knows the current page** — built from the ≤50 displayed rows, so you cannot filter by anyone not on page 1.
3. **`mcp` is an offered source no code path produces** — selecting it can only return zero rows.
4. **The page has zero translations.** `shared/i18n/{en,de}.json` contain no `admin.auditLog.*` keys; the German UI is entirely English.
5. **`docs/admin-ui.md:175` documents a feature that does not exist** — row expansion showing a diff.
6. **`server/tests/*.test.js` never run in CI** — the jest `testMatch` covers only `tests/unit/**` and `tests/integration/**`.

---

## Agreed design

**1. Fix the URL-state loss.** New `useFilterParams` hook batching a filter change and the page reset into **one** `setSearchParams` call. `useFilterState` stays untouched (it is correct for the single-param usage on 13 other pages) but gains a JSDoc warning.

**2. Checkbox multi-selects.** New reusable `FilterMultiSelect` for Resource, Action, Result, Source and Actor: a checkbox per option **with its entry count**, plus Select all / Select none, type-ahead above 10 options, full keyboard and ARIA support. Dates stay single-value.

**3. URL semantics — include set minus exclude set** *(confirmed by @manzke, [comment](https://github.com/intrafind/ihub-apps/issues/2213#issuecomment-5394649237))*

Each field gets two parameters: `<field>` (the include set, defaults to `*`) and `<field>Exclude` (subtracted from it). One matcher rule, exclusion wins:

```js
const matches = value =>
  (include.has('*') || include.has(value)) && !(exclude.has('*') || exclude.has(value));
```

| URL | Result |
|---|---|
| *(neither parameter)* | everything — the default |
| `?action=*` | everything — explicit form of the default |
| `?action=create,update` | only those two |
| `?actionExclude=login,logout` | everything except those two |
| `?actionExclude=*` | nothing — the "select none" state |
| `?action=*&actionExclude=login` | everything except login |
| `?resource=app&resourceExclude=app` | nothing — exclusion wins on a value in both |
| `?resource=app` | unchanged from today — existing links keep working |

For a *partial* selection the checkbox UI emits the **exclusion** form, so a resource type added by a later release stays visible in bookmarked views instead of being silently filtered out; inclusion remains fully supported on read. `actor` uses repeated parameters only and is never comma-split, since a username can contain a comma.

**4. Facets replace the hardcoded lists.** `GET /api/admin/audit-log?facets=1` returns `{value, count}` per field, computed in the **same file scan** as the query and over the date range only — so unticking `login` does not make the `login` checkbox disappear. This fixes defects 1 and 2 at once, and the counts are what make the original complaint tractable: you see `login — 794` and know exactly what to untick.

**5. Bounded perf fix.** `queryAuditLog()` currently parses every line of every file in the range into one array, sorts it all, then filters. Change to a single streaming pass — parse, count facets, filter, keep only matches.

**6.** i18n keys for en + de, `docs/admin-ui.md` rewrite (including correcting defect 5), changelog entries under `docs/releases/5.5.0/`.

**Tests:** a client regression test that fails against `main` today (toggling a filter must land in the URL *and* in the refetch query string), plus server tests covering the full include/exclude/`*` matrix, exclusion precedence, legacy `admin`-field actors, facet counts, and an actor named `Doe, John` matching as one value — all under `tests/unit/**` so they actually run.

---

## Status

- 📋 Full plan, decisions table, assumptions and risks: [`concepts/2026-08-24 Audit Log Filter Fix and Multi-Select Filters.md`](``https://github.com/intrafind/ihub-apps/blob/claude/issue-2213-planning-72lbxm/concepts/2026-08-24%20Audit%20Log%20Filter%20Fix%20and%20Multi-Select%20Filters.md)``
- 🔀 Planning PR: #2214 (draft, CI green)
- ✅ Blocking decision (URL semantics) resolved — implementation can start

**Still open, not blocking:**
- **Free-text search** over `summary` / `resourceId` / `ip` / `requestId` — not in the original report, but would often beat any combination of checkboxes. In scope?
- **Quick presets** — a one-click "Hide login/logout", plus "Failures only" / "Last 24h" chips.
- **Largest realistic audit volume?** If any installation writes millions of entries per week, a per-day facet cache would be needed instead of a live scan.
- **The `mcp` source** (defect 3) — wire it into the MCP path, or remove it from the UI?
- **Default date range** is 7 days — keep, or shorten to 24h now that filters work?
- **Orphaned `server/tests/*.test.js`** (defect 6) — fix the jest `testMatch` here or separately?

manzke:
Closes #2213. Implements the plan in [`concepts/2026-08-24 Audit Log Filter Fix and Multi-Select Filters.md`](``https://github.com/intrafind/ihub-apps/blob/claude/github-issue-2213-ovb9a3/concepts/2026-08-24%20Audit%20Log%20Filter%20Fix%20and%20Multi-Select%20Filters.md)`` (merged as #2214), with every open question answered per [this comment](https://github.com/intrafind/ihub-apps/issues/2213#issuecomment-5395268818). One PR, as requested.

## The bug

`handleFilterChange` called two `useFilterState` setters in the same tick:

```js
setter(value);      // setSearchParams(fn) -> navigate("?resource=app")
setPageParam('1');  // setSearchParams(fn) -> navigate("?")   <-- overwrites
```

React Router's `setSearchParams` does not queue the way `setState` does — its functional updater receives the **render-time** params — so the second call recomputed from the URL as it was *before* the filter, deleted `page` (it equalled its default), and navigated. The filter never reached the URL or the server.

This is no longer an inference: `tests/unit/client/use-filter-params.test.jsx` demonstrates it directly, and it stays in the suite as an executable warning so nobody removes the batched setter.

**`DataTable` had the same shape.** In server mode `handlePageSizeChange` called `onPageSizeChange(next)` **and** `onPageChange(1)` back to back — which is why changing rows-per-page reverted to 50, and why fixing only the page was not enough. DataTable now calls `onPageSizeChange` only; resetting the page is the consumer's job, documented on `normalizePagination`. The audit log is the client's only server-paged table, so nothing else is affected.

## The fix

New `useFilterParams` hook batches several parameters into **one** `setSearchParams` call. `useFilterState` is untouched — it is correct for the single-parameter usage on the other 13 pages — but gains a JSDoc warning about the trap.

## Checkbox filters with entry counts

Actor, resource, action, result and source are now checkbox lists (`FilterMultiSelect`, exported from `data-table/` for other admin tables): a checkbox per value **with its entry count**, Select all / Select none, type-ahead above ten options, arrow-key navigation and Escape to close. The counts are what make the original complaint tractable — you see `login — 794` and know exactly what to untick.

Options come from server-computed facets (`?facets=1`), counted in the **same file scan** as the query and over the **date range only**, so unticking a value never makes its checkbox vanish. That replaces two provably wrong lists: the hardcoded `RESOURCE_TYPES` (which offered `provider` — nothing writes it — and omitted `tool`, `credential`, `integrations`, `uiConfig` and every path-derived resource) and the actor list, which only knew the ≤50 rows on screen.

`mcp` is gone from the source filter — no code path writes it, so selecting it could only ever return an empty table. Filed as #2216 rather than removed from the schema, so wiring it up later needs no migration.

## URL contract

Each field takes an include parameter and a `<field>Exclude` parameter subtracted from it, `*` meaning "every value", exclusion winning:

```js
matches = value =>
  (include.has('*') || include.has(value)) && !(exclude.has('*') || exclude.has(value));
```

| URL | Result |
|---|---|
| *(neither)* | everything — the default |
| `?action=create,update` | only those two |
| `?actionExclude=login,logout` | everything except those two |
| `?actionExclude=*` | nothing — the "select none" state |
| `?action=*&actionExclude=login` | everything except login |
| `?resource=app&resourceExclude=app` | nothing — exclusion wins |
| `?resource=app` | unchanged from today — existing links keep working |

The checkbox UI writes the **exclusion** form for a partial selection, so a resource type added by a later release stays visible in a bookmarked view. `actor` reads repeated parameters only and is never comma-split (a username can be `Doe, John`); the other four accept both forms. A per-field value cap (500) and a search-term length cap (200) answer **400** rather than returning a silently wrong result.

## Also in scope

- **Free-text `?q=`** over summary, resource ID, IP, request ID and actor name — in-memory, applied in the same pass as the value filters.
- **Quick filters:** Last 24 hours / 7 days / 30 days, **Hide sign-ins**, **Failures only**, **Clear all filters** (which also covers Q10's "active filters" affordance).
- **Default range is 24 hours** on the page. `queryAuditLog`'s fallback for a caller sending *no* range stays at 7 days — `useOverviewData`'s `/admin/audit-log?limit=8` depends on it, and shortening it there would empty the admin overview's activity panel on a quiet installation. (The plan had assumed that call was unaffected; it would not have been.)
- **Single-pass scan.** `queryAuditLog` streamed each daily file once — parse, count facets, filter, keep matches — instead of parsing the entire range into one array and sorting all of it.
- **CSV export** shares the query parser, so an export always matches the table.
- **Full `admin.auditLog.*` translations** for `en` and `de`. The page previously had none, so the German UI was entirely English.
- **`docs/admin-ui.md`** rewritten, including dropping the row-expansion-with-diff view it promised and that has never existed (Q5).
- **Changelog** entries under `docs/releases/5.5.0/`.

## Files

| File | Change |
|---|---|
| `client/.../hooks/useFilterState.js` | add `useFilterParams`; document the single-setter-per-tick rule |
| `client/.../pages/AdminAuditLogPage.jsx` | batched URL updates, multi-selects, facets, search, presets |
| `client/.../data-table/filters/FilterMultiSelect.jsx` | **new** — reusable checkbox filter |
| `client/.../utils/auditLogFilters.js` | **new** — include/exclude encoding |
| `client/.../data-table/DataTable.jsx`, `useTablePagination.js` | one navigation per page-size change |
| `server/services/AuditLogService.js` | array + exclude matching, facets, free-text, streaming scan |
| `server/routes/admin/auditLogQueryParams.js` | **new** — shared query parser, testable without Express |
| `server/routes/admin/auditLog.js` | use the parser for list and export; `facets=1` |
| `shared/i18n/{en,de}.json` | `admin.auditLog.*` + `admin.filters.multiSelect.*` |

## Verification

`npm run lint:fix`, `npm run format:fix`, `npm run test:unit` → **415 tests, 38 suites, all passing**, including four new suites (75 tests):

- `tests/unit/server/auditLogQuery.test.js` — the full include/exclude/`*` matrix, exclusion precedence, unknown-value exclusion as a no-op, legacy `admin`-field actors, `Doe, John` matching as one value, facet counts and their independence from the active filters and from `q`, date-range scoping, pagination over the filtered set, malformed lines skipped.
- `tests/unit/server/auditLogQueryParams.test.js` — comma splitting, the actor exception, both caps, flag and integer clamping.
- `tests/unit/client/admin-audit-log-filters.test.jsx` — **the reported bug end to end**: a toggled filter must land in the URL *and* in the refetch query string. Plus page reset, page size, select all/none, legacy and exclusion links, presets, search, and keyboard access.
- `tests/unit/client/use-filter-params.test.jsx` — the root cause, pinned.

Also verified against a running instance (local admin login, seeded login/failure events): facets return real counts, every row of the URL table above behaves as specified, both caps answer 400, `limit` clamps, and the CSV export honours the filters.

Two follow-ups filed as agreed: #2216 (`mcp` source never written) and #2217 (`server/tests/*.test.js` never run in CI).

https://claude.ai/code/session_01SoWjWx5AiyaXLz6MCNm9aa

---
_Generated by [Claude Code](https://claude.ai/code/session_01SoWjWx5AiyaXLz6MCNm9aa)_

### DIFF
--- a/client/src/features/admin/components/data-table/DataTable.jsx
+++ b/client/src/features/admin/components/data-table/DataTable.jsx
@@ -96,8 +96,11 @@ function DataTable({
   };
   const handlePageSizeChange = next => {
     if (isServerPaged) {
+      // Only one callback, deliberately: a server-paged page usually keeps both
+      // values in the URL, and two navigations in the same tick lose the first
+      // (React Router's setSearchParams does not queue). Resetting to page 1 is
+      // therefore the consumer's job, inside its own onPageSizeChange.
       paginationConfig.onPageSizeChange && paginationConfig.onPageSizeChange(next);
-      paginationConfig.onPageChange(1);
     } else {
       setInternalPageSize(next);
       setInternalPage(1);
--- a/client/src/features/admin/components/data-table/filters/FilterMultiSelect.jsx
+++ b/client/src/features/admin/components/data-table/filters/FilterMultiSelect.jsx
@@ -1,0 +1,264 @@
+import { useCallback, useEffect, useId, useMemo, useRef, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import Icon from '../../../../../shared/components/Icon';
+
+// Above this many options the popover grows a type-ahead box.
+const SEARCH_THRESHOLD = 10;
+
+/**
+ * FilterMultiSelect — checkbox filter for a table's filter row.
+ *
+ * A trigger button showing how much of the vocabulary is currently selected,
+ * opening a popover with one checkbox per option (each with its entry count),
+ * plus **Select all** / **Select none** and, for long lists, a type-ahead.
+ *
+ * The control is fully controlled: it never derives state from its own
+ * interactions. `selected` is the list of currently selected option values and
+ * `onChange` receives the complete next list, so the page decides how that maps
+ * onto its URL parameters.
+ *
+ * Options are usually server-computed facets, so an option's `count` tells the
+ * admin what a checkbox is worth before they untick it.
+ *
+ * @param {object} props
+ * @param {string} props.label - Field label rendered above the trigger
+ * @param {Array<{value: string, label?: string, count?: number}>} props.options - Available values
+ * @param {string[]} props.selected - Currently selected option values
+ * @param {(next: string[]) => void} props.onChange - Receives the full next selection
+ * @param {string} [props.allLabel] - Trigger text when everything is selected
+ * @param {string} [props.noneLabel] - Trigger text when nothing is selected
+ * @param {string} [props.searchPlaceholder] - Type-ahead placeholder
+ * @param {string} [props.emptyLabel] - Shown when there are no options at all
+ * @param {boolean} [props.disabled=false]
+ * @param {string} [props.className]
+ */
+function FilterMultiSelect({
+  label,
+  options = [],
+  selected = [],
+  onChange,
+  allLabel,
+  noneLabel,
+  searchPlaceholder,
+  emptyLabel,
+  disabled = false,
+  className = ''
+}) {
+  const { t } = useTranslation();
+  const baseId = useId();
+  const containerRef = useRef(null);
+  const triggerRef = useRef(null);
+  const listRef = useRef(null);
+  const searchRef = useRef(null);
+
+  const [open, setOpen] = useState(false);
+  const [term, setTerm] = useState('');
+
+  const selectedSet = useMemo(() => new Set(selected), [selected]);
+  const allSelected = options.length > 0 && selectedSet.size >= options.length;
+  const noneSelected = selectedSet.size === 0;
+
+  const showSearch = options.length > SEARCH_THRESHOLD;
+  const visibleOptions = useMemo(() => {
+    const needle = term.trim().toLowerCase();
+    if (!needle) return options;
+    return options.filter(o =>
+      String(o.label ?? o.value)
+        .toLowerCase()
+        .includes(needle)
+    );
+  }, [options, term]);
+
+  const close = useCallback(() => {
+    setOpen(false);
+    setTerm('');
+    triggerRef.current?.focus();
+  }, []);
+
+  // Dismiss on click outside. Registered only while open.
+  useEffect(() => {
+    if (!open) return undefined;
+    const onPointerDown = event => {
+      if (!containerRef.current?.contains(event.target)) {
+        setOpen(false);
+        setTerm('');
+      }
+    };
+    document.addEventListener('mousedown', onPointerDown);
+    return () => document.removeEventListener('mousedown', onPointerDown);
+  }, [open]);
+
+  // Move focus into the popover so the keyboard path works without a mouse.
+  useEffect(() => {
+    if (!open) return;
+    const target = showSearch
+      ? searchRef.current
+      : listRef.current?.querySelector('input[type="checkbox"]');
+    target?.focus();
+  }, [open, showSearch]);
+
+  const toggleValue = value => {
+    const next = new Set(selectedSet);
+    if (next.has(value)) next.delete(value);
+    else next.add(value);
+    // Preserve the option order rather than the click order, so the emitted
+    // list is stable and comparable.
+    onChange?.(options.filter(o => next.has(o.value)).map(o => o.value));
+  };
+
+  // Arrow keys move between checkboxes, Escape closes and returns focus. Bound
+  // natively on the container while open so it catches keys from the
+  // type-ahead box and the select-all/none buttons as well.
+  useEffect(() => {
+    if (!open) return undefined;
+    const container = containerRef.current;
+    if (!container) return undefined;
+    const onKeyDown = event => {
+      if (event.key === 'Escape') {
+        event.preventDefault();
+        close();
+        return;
+      }
+      if (event.key !== 'ArrowDown' && event.key !== 'ArrowUp') return;
+      const boxes = Array.from(
+        listRef.current?.querySelectorAll('input[type="checkbox"]:not([disabled])') ?? []
+      );
+      if (boxes.length === 0) return;
+      event.preventDefault();
+      const current = boxes.indexOf(document.activeElement);
+      const delta = event.key === 'ArrowDown' ? 1 : -1;
+      const next = current === -1 ? 0 : (current + delta + boxes.length) % boxes.length;
+      boxes[next].focus();
+    };
+    container.addEventListener('keydown', onKeyDown);
+    return () => container.removeEventListener('keydown', onKeyDown);
+  }, [open, close]);
+
+  const triggerText = allSelected
+    ? (allLabel ?? t('admin.filters.multiSelect.all', 'All'))
+    : noneSelected
+      ? (noneLabel ?? t('admin.filters.multiSelect.none', 'None'))
+      : t('admin.filters.multiSelect.someSelected', '{{selected}} of {{total}}', {
+          selected: selectedSet.size,
+          total: options.length
+        });
+
+  const filtered = !allSelected;
+
+  return (
+    <div className={className} ref={containerRef}>
+      {/* A plain span, not a <label>: the trigger's accessible name is built
+          from the field name *and* the button's own text via aria-labelledby,
+          so a screen reader announces "Action, 3 of 8". */}
+      <span
+        id={`${baseId}-label`}
+        className="block text-sm font-medium text-gray-700 dark:text-gray-300 mb-1"
+      >
+        {label}
+      </span>
+      <div className="relative">
+        <button
+          id={`${baseId}-trigger`}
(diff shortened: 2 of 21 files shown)


manzke:
All seven review findings confirmed against the code and fixed in c6b5098. Three involved a judgement call worth flagging:

**Filter-value cap (finding 2).** The suggestion was to encode whichever of the include/exclude sets is smaller. I've made the fallback conditional instead: the exclusion form is still preferred whenever it fits, and the inclusion form is used only when the exclusion list would exceed the cap. The two forms differ in intent for *future* values — an inclusion list pins an exact set, so a resource type added by a later release would be silently invisible in a bookmarked view, which is the wrong failure mode for an audit log (this was decided as D2a with @manzke). Always taking the smaller side would flip that in the common case; the conditional fallback only kicks in where the alternative is a 400.

The cap also moved 500 → 2000. It is hygiene, not a tight bound — matching a value is a Set lookup either way — and at 2000 a selection that fits neither form needs more than 4000 distinct values in the date range. A test now pins the client and server constants together, since a mismatch is exactly what produced the bug.

**"Last 24 hours" (finding 3).** Correct, and it was my mislabel — the date filter selects whole calendar days by filename with no timestamp cutoff. I chose relabelling over adding timestamp-level filtering because `from`/`to` are day-granular by contract (they're `<input type="date">`), so a rolling window would need a different filter shape. Presets are now **Today** / **Today & yesterday** / **Last 7 days** / **Last 30 days**, and the docs, changelog and default are described as calendar days. The default still spans two days rather than one so the table isn't near-empty just after midnight — which is what the short default was for.

**Result window (finding 1).** `queryAuditLog` now keeps only the newest `offset + limit` entries, trimming whenever the buffer reaches twice that, and counts `total` separately. Verified against a running instance that `limit=1&offset=N` returns exactly the Nth row of the unpaged list with an unchanged `total`.

The other four were straight fixes: the table's own action/result/source cells now go through `valueLabel` (they were rendering raw values, which made the "fully translated" claim false in the table itself), `pageSize` from the URL is clamped to the largest page size the table offers, `aria-haspopup` is gone since the trigger is a disclosure rather than a menu, and the page-reset test now starts at `?page=4` so it can actually fail.

New coverage: the result window (same page and same total at any window size, including zero), the encoder's over-cap fallback and its preference for exclusion when it fits, the page-size clamp, and the table rendering translated values. 430 tests / 39 suites green, lint and format clean.

---

PR #2226, opened 2026-08-25
Visual loop containers, parallel iteration, and friendlier workflow editing

manzke:
## Summary

Implements the Phase 1+2 recommendations from the merged RAGFlow comparison concept ([#2225](https://github.com/intrafind/ihub-apps/pull/2225)): loops become **visual containers** on the workflow canvas, iteration can run **in parallel** or **branch**, the editor gets the non-technical treatment, and every shipped workflow that hand-built iteration gets a modernized copy. Two later passes move the remaining loop plumbing onto the steps themselves and fix what building all of it exposed.

The headline user story now works end to end with **6 nodes and zero expressions**: search/outline → *drop an AI step inside a loop box* → report. Previously this required the 13-node cursor idiom with JSON-only configs.

Five commits: (1) loop containers + editor, (2) new-workflow start/end fix, (3) modernized v2 workflows + conditional/nested body support, (4) loop bookkeeping and progress notes as step options, (5) variable discovery, tool form and container layout fixes.

## Engine

- **Container loop bodies**: a `loop` node&#39;s body can be defined by child nodes carrying `parentId` — `LoopNodeExecutor` resolves them from the workflow definition and orders them topologically by their edges. Inline `config.body` keeps working unchanged (serializer-generated agent workflows are unaffected).
- **Conditional steps inside a body**: edges *between* body nodes are evaluated during each iteration, so an iteration can branch and skip steps. A step with no matching outgoing edge ends the iteration; a body with no edges runs every node in order (as before). A per-iteration step cap stops a cyclic body and logs a warning.
- **Nested containers**: a container may be a body node of another container. Loop variables are saved and restored around an inner loop, so steps after it still see the outer `_loopItem` (previously the inner loop deleted them).
- **Parallel forEach** (`concurrency: 1–10`): bounded worker pool over snapshots of the pre-loop state; results collected in item order; cross-iteration state writes deliberately not propagated (deterministic runs).
- **Pause guard**: a body step returning `paused` (human-in-the-loop) now fails with a clear message instead of silently running on — pause/resume is driven by the outer scheduler, which only sees the loop node.
- **Schema**: nodes accept `parentId` (must reference a loop node, cycle-safe), `size`, and `progress`; container children are exempt from the incoming-edge refinement; edges may not cross a container boundary. `DAGScheduler.findStartNodes/findEndNodes` exclude children.

## Step options replace loop plumbing

Reviewing the converted workflows showed that most of what a loop body contained was not the work — it was plumbing: a step to copy the current item under a usable name, a step to announce what was about to happen, a step to bump a counter. Those are now options on the steps that already exist.

- **`itemVariable` on a loop** names the current element, so body steps write `{{_currentDoc}}` instead of `{{_loopItem}}`. The name is loop-scoped like the four built-in loop variables, which is what makes nested loops legible — an inner loop no longer shadows the outer item for steps placed after it.
- **`countInto` on a loop** names a state path the loop increases by one after every finished round, and defines it as `0` before the first round so a `while` condition can read its own counter immediately. Works in every mode. Prototype-polluting paths are rejected.
- **`progress: { message, when?, status? }` on any node** announces the step just before it runs, replacing the separate announcement step that used to precede it. The optional `when` lets a step speak up only in the rounds where it has something to say. Notes show even when the step sets `chatVisible: false` — that pairing is the point: hide the mechanical step, show the one line that means something. Shared helper in `server/services/workflow/nodeProgress.js`, called from `WorkflowEngine` for top-level steps and `LoopNodeExecutor` for body steps; the standalone `progress` node keeps its plain-string config and now uses the same helper.
- **Variables you no longer define at all.** With counters seeded by their loop and `_records`/`_corpus`/`_coverage` created by the steps that own them, the &#34;initialise counters&#34; transform at the top of a workflow is unnecessary; the shipped examples drop theirs and the one genuine constant moves to the Start step&#39;s defaults.

Between the two passes the shipped examples lost most of their loop plumbing — the iFinder review&#39;s per-document loop went from 12 steps to 3 (fetch, extract, record) and its top level is 7 steps with none of them plumbing.

## Knowing which variables exist

Guessing a name was the easiest way to break a workflow quietly: a misspelled `{{topci}}` renders as an empty string and the run carries on.

- **Variables tab.** Each step&#39;s config panel lists every name in scope, grouped by origin — workflow inputs, each earlier step&#39;s output, the enclosing loops&#39; items and counters, and the engine&#39;s run metadata. Click to copy as `{{name}}`.
- **Unknown-name warning** on the canvas for any `{{name}}` or `$.data.name` no step defines, clicking through to the step. It knows what steps write beyond `outputVariable` (a corpus search&#39;s `corpusVar`/`coverageVar`, a loop&#39;s `itemVariable`/`countInto`, a transform&#39;s operation targets, a start step&#39;s `inputVariables`/`inputMapping`/`defaults`, a verifier&#39;s `verificationResult`) and the scopes some steps render against (a report template&#39;s `records`/`coverage`/`synthesis`, a prompt&#39;s `{{sources}}` placeholder). Handlebars block bodies are skipped — names there are block-relative and flagging them would make the check untrustworthy — but the block&#39;s own argument is checked.

It caught a bug in this PR&#39;s own edit, and one that predates it: **`iterative-research-human`&#39;s feedback loop has never worked.** The engine keyed a checkpoint answer as `humanResponse_plan-checkpoint`, which no template can reference, while the prompts read `humanResponse_plan_checkpoint`. Human steps now honour `outputVariable` like every other step, and that workflow names its two answers.

## Editor

- Loop nodes render as **resizable dashed containers** with a plain-language header (&#34;for each item in *searchResults* · parallel ×2&#34;), an item/index hint chip, and a &#34;collects → analyses&#34; badge. **Dragging a node into the box adopts it as loop body; dragging it out releases it** — boundary-crossing edges are cleaned up with a notice. Containers can be nested by dragging one into another (never into itself or its own body).
- **Auto-layout now arranges loop bodies.** It previously skipped every node with a `parentId`, leaving body steps wherever they sat. It lays out each container&#39;s steps from their sibling edges, innermost container first, and grows each container to fit.
- **Tool steps are configurable again.** The form bound `toolName`, a key the engine never reads, so the picker always looked empty; Parameters bound an object to a plain textarea, rendering `[object Object]`, and the first keystroke would have replaced the parameter map with a string. It binds `toolId`, the picker offers callable function ids (`iFinder_getContent`, not just `iFinder`), and structured values use a new `JsonField` that keeps the last valid value while you type. `FormField` now refuses objects in textareas too, and HTTP headers use the new editor.
- **Reading agent memory is a real step.** `memory` was a palette entry with no executor — adding one would have failed at run time — while the capability hid inside a transform operation. It is now a node type with an executor and form, sharing one section slicer with the transform operation, degrading to an empty string when a profile or section is missing.
- **Searchable node palette** with icons, friendly localized names and hover descriptions; **all 24 node types** have palette entries and config forms (9 were previously JSON-only).
- **Decision nodes get labeled yes/no source handles** that auto-attach the matching edge condition; **clicking an edge opens a condition builder** with variable suggestions.
- **Plain-language LoopForm**: mode select, list picker fed by upstream `outputVariable`s, parallel option with caveat, max-iterations slider, plus *Name the current item* and *Count completed rounds into*. Every step&#39;s form has a **Progress note** field with an *only show when* condition.
- **Lossless canvas round-trip**: node descriptions, execution settings, non-English names, and edge handles/labels/conditions survive saving.
- **New workflows are valid on creation** (bug fix, present on `main` today): &#34;Create New Workflow&#34; started from an empty definition the schema always rejected. Both creation paths now start from one shared template with connected Start and End steps.
- Papercuts: dead &#34;Visual editor&#34; list action (`/editor` → `/edit`), `window.prompt()` → proper ID dialog, wrong expression placeholders corrected to real `$.data.` syntax, missing `verifier.mode` field.

## Integrations

- **iFinder `getContent` reports truncation** instead of leaving callers to re-derive it: always returns `truncated`, `originalContentLength`, `returnedContentLength` and the applied `maxLength`. The iFinder review reads the flag off the result, which is what let its per-document loop drop to three steps.

## Workflows &amp; docs

- New default workflow **`topic-deep-dive`**: outline a topic → analyze every section inside a parallel loop container → compose a report.
- **Modernized copies of every iterating workflow** (originals untouched, prompts preserved byte-for-byte; a node-by-node diff and a dangling-variable scan confirm only loop plumbing changed):
  - `stellungnahmen-review-ifinder-v2` — two containers (a `while` search-refinement loop counting its rounds via `countInto`, and a `forEach` per-document loop that announces each load and flags truncated ones with a `when`-guarded note)
  - `stellungnahmen-review-v2`, `corpus-analysis-direct-v2` — one `forEach` container each, two steps long
  - `corpus-analysis-decomposed-v2` — **nested** containers, per sub-question × per document
  - `iterative-research-auto-v2` — the think/research/refine cycle as one `while` container
  - `iterative-research-human` is deliberately **not** converted: its checkpoints pause from inside the iteration, which containers cannot support (see the pause guard).
- `docs/workflows.md` (registered in `docs/SUMMARY.md`) covering containers, conditional bodies, nesting, named items, round counting, progress notes, variable discovery and the parallel caveat; changelog entries in `docs/releases/5.5.0/`; i18n keys (en/de).

## Validation

- 559 unit tests (45 suites) pass, including new suites for node progress, shipped-workflow shape, container auto-layout, and variable references — the last asserting every shipped workflow resolves, so this stays fixed. All 18 workflows validate against the schema; a real server boot copies and loads the new defaults with zero errors.
- `lint:fix` / `format:fix` clean (0 errors); client production build passes.
- Two jsdom gaps surfaced and are polyfilled in the shared jest setup: `setImmediate` (winston) and `structuredClone` (dagre). A JSON round-trip is **not** a valid `structuredClone` — it drops `undefined` and Map/Set values and leaves dagre laying every node on top of the next; v8&#39;s serializer is used instead.
- One earlier test asserted wall-clock time for parallel iteration and proved flaky under load — replaced with a deterministic in-flight concurrency probe.

## Out of scope (later phases per the concept)

Canvas test-run/live overlay, version history UI, workflow template gallery, AI workflow generation, trigger UI. Also deferred: composite &#34;task&#34; nodes that fold a whole search-until-complete cycle into one configurable step — worth prototyping once this lands.

https://claude.ai/code/session_01Rt6CUNAr1bTzrZmMGKHTQw


manzke:
@claude review

manzke:
Reviewed the branch. Fifteen findings, all defects this PR introduced; the actionable ones are fixed in dd006db.

**One that mattered a lot: `countInto` never advanced.**

`bumpCounter` mutated `currentState`, and the very next line replaced it with the object `executeBodyNodes` had copied *before* the bump. In every sequential mode a top-level counter stayed at its seeded `0`. Nested paths like `_coverage.processed` only appeared to work because the shallow copy shares the nested object.

That left `stellungnahmen-review-ifinder-v2`'s search budget dead: its `while` condition and its body edge both read `_searchIterations`, so nothing but `maxIterations: 3` was bounding the LLM refinement rounds. The bump now happens after the loop adopts the body's state.

The three existing tests all missed it — one used a nested path, one always failed the body, one used parallel mode where `currentState` is never reassigned. There are now tests for a flat counter and for a `while` loop that bounds itself on its own counter, and I confirmed both fail against the old code.

**Loop bodies silently dropped steps.** `executeBodyNodes` started only at `bodyNodes[0]` whenever any sibling edge existed, so a second entry point or an unwired step never ran — contradicting `resolveContainerBody`'s own documented guarantee. It also followed only the *first* matching outgoing edge, so a fan-out inside a loop ran one branch while the identical graph at top level runs both. Entry points are now every child with no incoming sibling edge, every matching edge is followed, and a step runs at most once per iteration so a diamond rejoins instead of double-executing.

**Nested containers were mishandled in the editor.** `parentsFirst` only split parented from unparented nodes, so a grandchild could precede its own parent and React Flow could not resolve it. `containerAt` returned the first match in array order — always the *outer* container — so dropping a step into an inner loop silently attached it to the outer one, where it would run once per sub-question instead of once per document. Both fixed; nesting is a headline feature of this PR, so it is now covered by tests.

**Publishing a new workflow posted to `/workflows/new/publish`** — replacing the blocking `window.prompt` with an async dialog left `handlePublish` running ahead of it. **Edge conditions compared strings under strict equality**, so "Compare with `3`" could never match the number `3`; the builder now types clean literals and states how a value will be compared.

Also fixed: LoopForm displayed `forEach` while the engine defaults to `for`; widening `FormField` had made object HTTP bodies read-only; `StructuredRecordForm` held schema text in local state that never re-seeded, so switching nodes could write one node's schema onto another (every node form is now keyed by node id, so no form can carry a stale value across a selection change); the boundary-edges-removed notice was overwritten in the same tick, making lost wiring silent; `handleDeleteNode` called other setters inside a `setNodes` updater; blank lines in corpus search queries reached the backend as empty strings; the reference checker missed the transform `merge`/`into` target and flagged a correct step; and twenty `t()` keys this PR added were in neither locale file.

574 unit tests green, all 18 workflows validate, client builds. The migration-lock errors on a first cluster boot reproduce with these changes stashed, so they are not from this branch.

---

manzke:
Both bot reviews ran against `846f813`, so eight of their findings — `countInto` being discarded, `parentsFirst` nesting, `containerAt` picking the outer container, form state going stale across selection, the missing Variables-tab translations, publish firing before save, body-walk entry points, and the swallowed boundary-edge notice — were already fixed in dd006db. The rest are addressed in e9ac21e.

**Parallel iteration wasn't delivering its snapshot guarantee.** Each worker got a *spread* of pre-loop state, so nested objects and arrays stayed shared — a body step writing through one leaked into every other iteration, precisely what the snapshot exists to prevent. Workers now get a per-worker deep copy, with engine-internal keys carried by reference since the engine mutates those after the executor returns. The parallel path was also handed the pre-loop state rather than the copy `seedCounter` had already initialised, so a body step reading its own `countInto` path saw `undefined` where a sequential run sees `0`.

**A pausing body step reported success.** It was recorded as a failed iteration and then returned as a completed loop, so the engine scheduled the rest of the workflow and the guard's message never surfaced — which defeated the guard's whole purpose. A body result carrying an error now fails the loop. Ordinary iteration failures keep their existing tolerated behaviour: that predates this branch, and changing it would alter how existing workflows handle a failing round, so only an errored step (today just the pause guard) ends the run.

**Loop item names were global.** `itemVariable` was registered as a workflow-wide producer, so a step *after* the loop was offered a variable that no longer exists by then, and no warning fired when it read one. It is now contributed only to steps inside its own loop, walking the enclosing containers; `countInto` stays global because it outlives the loop. The producer list also read `outputVariable` for query-plan steps, which use `outputVar`, and missed template-render's `reportVar` — both made the checker flag correct steps, which is the failure mode that makes a checker worth ignoring.

**MemoryForm** decided its mode on whether the profile field held a value, so choosing "a fixed agent profile" wrote `''` and the form flipped straight back before anything could be typed.

**One I did not act on:** the palette labels and form field labels being English rather than translation keys. That is the existing convention in this directory — form labels were plain strings on `main` before this branch — so translating the palette alone would leave the editor *more* mixed, not less. It's worth doing as its own change across all workflow-editor forms, and I'm happy to take that on separately if you want it.

585 unit tests green. Every new test was checked against the unfixed code first — the parallel isolation, parallel seeding, and loop-scope tests each fail without their fix.

---

PR #2246, opened 2026-08-28
feat: user-managed personal API keys in the integration section

### LINKED ISSUE #2232: User-specific OAuth API Key in the integration section
We should extend the integration section for users, where they can connect certain integrations. In this section a user should also be able to generate easily an API key. 
If this integration is available, with what settings, etc. must be configured by an admin.

So a user just clicks it and gets the information about urls and api keys/client id and secret.

manzke:
Closes #2232.

## What this adds

**Settings → Integrations** gets a **Personal API Key** card. A user clicks once and gets
credentials that let external tools call iHub Apps as them, plus the endpoint URLs those
credentials work against — no ticket to an administrator, no shared service account.

A personal key **acts as its owner**: calls made with it see exactly the apps, models and prompts
that user sees in the web UI, and never get admin access, even when the owner is an administrator.

Two credentials come from the same key:

| Credential | What it is | Where it's used |
| --- | --- | --- |
| **API key** | long-lived bearer token | `Authorization: Bearer …` on any call |
| **Client ID + secret** | OAuth 2.0 client credentials | `POST /api/oauth/token` for short-lived tokens |

Both are shown exactly once, when the key is created or rotated.

## Design

Each key is backed by an ordinary entry in `oauth-clients.json` marked `personal: true` and
carrying an owner snapshot. Reusing the client store rather than adding a second credential store
means rotation, suspension, cluster cache invalidation, admin visibility and audit logging all come
for free, and administrators keep one place to look for "who can call this server".

Tokens use a new auth mode, `oauth_personal_key`, treated exactly like an authorization-code token:
subject is the user, permissions resolve from that user's groups on every request, admin access is
denied unconditionally.

**Identity is resolved, not frozen.** The token carries only `sub`, `client_id`, `client_name`,
`scopes`, `jti` and `key_generation`. Name, e-mail and groups come from the client record on each
request — so revoking a key, suspending it, or turning the feature off cuts access immediately
rather than at expiry. The one snapshot is group membership (written on create, refreshed on
rotate), because a live lookup is impossible for OIDC and proxy users; the authorization-code flow
makes the same trade-off, and the docs say so plainly.

**Rotation is enforced by a counter, not a timestamp.** Each client carries
`metadata.keyGeneration`; every issue advances it, both minting paths stamp it into the token, and
both authentication paths check it — so rotating a key also invalidates the access tokens already
exchanged from its previous client secret. `iat` cannot do this job: it has second granularity, so a
credential minted in the same second as the rotation that replaced it is indistinguishable by time.

The comparison is `>=`, not equality. iHub runs clustered and each worker serves the client store
from its own cache that a cluster announcement refreshes asynchronously, so the request that mints a
credential writes the new generation on one worker while the request that first uses it may land on
another that has not caught up. Under equality that is a rejected key on its first use — measured
against a four-worker cluster, 13 of 15 rotate-then-use sequences failed. With `>=` a credential is
refused only for carrying a *lower* generation than the client, which is exactly the superseded
case.

## Administrator controls

`platform.oauth.personalKeys`, seeded by migration `V084`, all conservative:

| Field | Default | Purpose |
| --- | --- | --- |
| `enabled` | `false` | offer the feature at all |
| `allowedGroups` | `[]` | restrict who may create keys (empty = every signed-in user) |
| `maxKeysPerUser` | `5` | cap concurrent keys per user |
| `defaultExpirationDays` | `90` | lifetime when the user does not choose one |
| `maxExpirationDays` | `365` | upper bound the user cannot exceed |
| `allowClientCredentials` | `true` | also issue a client ID and secret |
| `scopes` | `[]` | scopes per key (empty = `mcpServer.defaultScopes`) |

Exposed under **Admin → OAuth → Authorization Server**. The feature also requires
`oauth.enabled.clients`; the page says so when the toggle is on but clients are off.

Both switches bind keys that already exist, not only ones created afterwards: turning `enabled` off
stops every key at once, and turning `allowClientCredentials` off stops existing client secrets at
the token endpoint while leaving the API keys working.

## Security model

- **A key never exceeds its owner.** Permissions resolve from the owner's groups on every request.
- **A key never reaches the admin API.** `/api/admin/*` is closed to personal keys and to every
  other delegated or machine principal, even when the owner is an administrator.
- **A key cannot mint another key.** Key management accepts only an interactive session: an OAuth
  service account, a static API key, a delegated authorization-code token, a personal key and an
  agent principal are all refused. Otherwise a narrow, short-lived delegation could be traded for a
  long-lived key carrying the owner's full permissions.
- **A key cannot outlive its policy.** The lifetime is capped by `maxExpirationDays` and bounds both
  credentials — the API key through its own `exp`, and the client credentials at the token endpoint
  and on each request, so the client ID and secret cannot keep minting owner-bound tokens after the
  key has expired.
- **Users only see their own keys.** A foreign key is reported as 404, never 403.

## Security fix included

End-to-end testing surfaced an escalation path this feature would have widened: `adminAuth` and
`contentAdminAuth` resolved admin rights from **raw group membership**, ignoring
`enhanceUserWithPermissions`, which already forces `isAdmin: false` for delegated and machine
principals. Any token carrying an admin's groups therefore reached `/api/admin/*` — already true
for OAuth service accounts and authorization-code tokens; this feature would have made it a
one-click path.

`isAdminEligiblePrincipal()` now states the rule once in `server/utils/authorization.js`, and both
middlewares consult it before looking at groups. It denies OAuth client credentials, static API
keys, authorization-code tokens, personal API keys and agent principals. Browser sessions and
non-admin APIs are untouched.

## Review round

An automated review raised 18 findings; all are addressed, and
[this comment](https://github.com/intrafind/ihub-apps/pull/2246#issuecomment-5494837383) explains
the three places where the fix differs from the literal suggestion. Beyond the rotation counter and
the expiry enforcement above, that round also brought: `grantTypes` reconciled on every issue, so a
key created while client credentials were disallowed no longer keeps an empty grant list and then
shows a secret the token endpoint always rejects; per-owner serialization of the count-then-insert,
so concurrent requests can no longer all pass `maxKeysPerUser`; `validateIdForPath` on rotate and
delete; `lastUsed` recorded on the HTTP and MCP paths rather than only at the token endpoint; a
failed list refresh no longer discarding a create or rotate response that carries one-time secrets;
a clipboard guard where the API is absent; switch semantics on both new admin toggles; and a cap on
the generated key name so a long user name cannot produce a client ID that later fails path
validation, which would have left the key unrotatable and unrevokable.

CodeQL then flagged two prototype-polluting assignments in the new store write. The key id arrives
from the request path, so the lookup now uses `Object.hasOwn` — the guard the rest of the client
store already uses — with a test that `__proto__`, `constructor` and `prototype` are refused
without touching `Object.prototype`.

## Verification

`npm run test:quick` passes (exit 0): 48 suites, 611 tests, plus
`server/tests/personalApiKeys.test.js` (92 checks) which is wired into that chain. The integration
suite is 91 passed. The client bundle builds. Lint and Prettier are clean on every changed file
(the remaining warnings are pre-existing).

Verified end to end against a running **four-worker cluster** with OAuth clients, the MCP gateway
and personal keys enabled:

- key creation returns the key, client ID and secret; the token's `sub` is the owner and it carries
  no frozen group list
- the key authenticates against `/api/inference/v1/models` (200) and `/mcp` initialize (200)
- `client_credentials` exchange yields an owner-bound token with `authMode: oauth_personal_key`
- **admin API via the key → 403**, content-admin API via the key → 403, service-account token → 403,
  while the admin's browser session still gets 200
- the key cannot mint another key → 403
- rotation invalidates both credential types: old API key → 401 on the HTTP API and on MCP, the
  access token already exchanged from the old secret → 401, old client secret → `invalid_client`;
  the new API key and secret work
- expiry bounds the client credentials: exchange → `invalid_client`, an access token minted earlier
  → 401, and rotating revives the key
- both administrator switches apply without a restart, and the new fields survive the platform
  config validator
- 20 consecutive rotate-then-use sequences with no failures (13 of 15 failed before the `>=` fix)
- a traversal key id on rotate and delete → 400

## Known adjacent issue, deliberately not changed

`jwtAuth` still compares `decoded.iat * 1000` against `client.lastRotated` for
`oauth_client_credentials` and `oauth_static_api_key` tokens, which has the mirror-image bug: a
token issued in the same second as a rotation can be rejected on its first use. Fixing it would
change auth behaviour for existing service accounts, so it is left alone and noted in the concept
document.

Also noted there: the per-owner serialization that makes `maxKeysPerUser` reliable is in-process.
Across cluster workers `oauth-clients.json` remains last-write-wins, as it is for every other write
to it, so two workers admitting a key at the same instant could still exceed the cap by one. Making
the store transactional would change every caller and belongs in its own change.

## Docs

- `docs/personal-api-keys.md` (new, linked from `docs/SUMMARY.md`)
- `docs/releases/5.5.0/features.md` — the feature; `fixes.md` — the admin-gate fix
- `concepts/2026-08-28 User-Specific OAuth API Keys in the Integration Section.md` — design note,
  including why rotation is a counter and what the cluster cache does and does not guarantee

---

https://claude.ai/code/session_01DdHeW72NkRki2v4DNoyqns

---
_Generated by [Claude Code](https://claude.ai/code/session_01DdHeW72NkRki2v4DNoyqns)_


manzke:
## CI status: one red check, `CodeQL` — not this PR's defect

Everything else on `2192c32` is green: `test-quick`, `test-integration`, `Build Check`, `auto-lint-format`, `axe-core`, `label`, and both CodeQL `Analyze` jobs. The only failure is the aggregated **`CodeQL`** alert gate: *"11 new alerts including 11 high severity security vulnerabilities"*.

### What the 11 alerts actually are

All 11 are the same query — **`js/insufficient-password-hash`**, *"Password from a call to `loadOAuthClients` is hashed insecurely"* — and every one of them lands in a file this PR does not touch:

| File | Lines | What the flagged code does |
| --- | --- | --- |
| `server/configCache.js` | 550 | `generateETag()` — sha256 over the config |
| `server/configCache.js` | 1055, 1643, 1679, 1724, 1777 | md5 over sorted `workflowIds` / `appIds` / `modelIds` / `toolIds` / `skillNames` for user-specific ETags |
| `server/routes/chat/dataRoutes.js` | 420 | md5 over sorted `promptIds` for a user-specific ETag |
| `server/routes/themeRoutes.js` | 188 | `generateETag()` — md5 over the theme config |
| `server/services/pwa/PwaService.js` | 43 | `computePwaETag()` — sha1 over the pwa config |
| `server/utils/chromeExtensionId.js` | 50, 140 | sha256 over an SPKI-DER **public** key (CRX3 extension ID, per spec) |

Not one of these is a password hash. They are ETags, cache-busting content hashes, and a CRX3 extension ID. `server/configCache.js:550` already carries `// lgtm[js/insufficient-password-hash] -- ETag, not a password hash`, so this exact false positive has fired here before — the legacy `lgtm[…]` annotation simply isn't honoured by CodeQL's default setup.

### Why it's attributed to this PR

CodeQL taints the object returned by `loadOAuthClients()` as a "password" because it carries a `clientSecret` property (a bcrypt hash). That object reaches the shared config cache, and from there the ETag helpers above.

That linkage is entirely pre-existing — on `main`, before this branch, `loadOAuthClients()` and `saveOAuthClients()` already called `configCache.get()` / `configCache.setCacheEntry()` (lines 38, 65, 89 and 146 of `server/utils/oauthClientManager.js` at `25f528a`). This PR only adds new *call sites* of those two functions, which places changed lines on the existing data-flow path — so the same pre-existing sinks get re-reported with a new path and are counted as "new in code changed by this pull request".

### What I did and did not do

There is no fix to port: the sinks are unrelated ETag code, and adding suppressions across five untouched files would widen this PR without working (inline CodeQL suppression is not supported for JS in default setup). I have not re-run the check either — CodeQL is deterministic static analysis, so a re-run would return exactly the same 11 alerts.

Two options for whoever reviews this, neither of which belongs in this PR:

1. **Dismiss the 11 alerts as false positives** in the Security tab. Lowest effort, and accurate.
2. **Break the taint at the source** — keep bcrypt-hashed client secrets out of the globally shared config cache, so the OAuth client store stops feeding every ETag computation in the app. That is a genuine hardening of `oauthClientManager` and would retire this false positive permanently, but it changes how client records propagate across cluster workers and is well outside the scope of this change. Happy to open it as a separate PR if you want it.

I'm keeping this PR watched until it is merged or closed.

---

manzke:
All 18 review findings are addressed in 925028b. Three of them I implemented differently from the literal suggestion, and one had a premise worth correcting — flagging those here so nothing looks quietly skipped.

**Rotation generation is compared with `>=`, not equality.** The finding is right that `iat` cannot express "issued before this rotation" and that `jti` does not help, so each personal client now carries `metadata.keyGeneration`, both minting paths stamp it into the token, and both authentication paths check it. But exact equality is not safe here: iHub runs clustered, and each worker serves the client store from its own cache that a cluster announcement refreshes asynchronously. The request that mints a credential writes the new generation on one worker while the request that first uses it may land on another that has not caught up yet. Measured against a four-worker cluster, **13 of 15 rotate-then-use sequences rejected the key that had just been issued.** With `>=`, a credential is refused only for carrying a *lower* generation than the client — exactly the superseded case — and a lagging reader can never reject something newer than what it knows about. The relaxation is that during the lag window a just-superseded credential can still be accepted by a worker that has not seen the rotation; that is the same window in which a deleted or suspended client is also still accepted, so it is a property of the store's cache rather than of this check.

**The key's stored expiry is not re-checked for the API key itself.** It is enforced at the token endpoint (the primary gate) and on each request for tokens exchanged from the client credentials — which is the actual gap the findings identified, since those have a short lifetime of their own that can outlast the key. The API key, though, is a JWT whose `exp` is set from the same lifetime and is already verified on every request; `metadata.apiKeyExpiresAt` is written from that same value, so it is a mirror rather than an independent bound. Re-checking it would add no enforcement and would reintroduce the cache-lag window above.

**`maxKeysPerUser` is serialized per owner, not made transactional.** Counting and inserting are now chained per owner, so concurrent requests in one worker can no longer all pass the check — verified by a test that fails without it (5 parallel creates against a limit of 2). A store-level transaction would change every caller of `oauth-clients.json`, so across workers the store stays last-write-wins as it is for every other write to it; two workers admitting a key at the same instant could still exceed the cap by one. That limitation is written down in the concept document rather than papered over.

**On the two additional authorization mocks:** added as asked, but the stated consequence could not actually occur — `server/tests/admin-endpoints-security.test.js` and `server/tests/authentication-security.test.js` both fail to load on `main` for an unrelated pre-existing reason (`jest is not defined`, they use `jest.mock` without importing it under ESM), so no request ever reaches the middleware there. The exports are in anyway so the mocks are correct when those suites are repaired.

Everything else was implemented as suggested, including the delegated-principal gate on key management (an authorization-code token, static API key, service account or agent principal can no longer mint a key), reconciling `grantTypes` on every issue, the `allowClientCredentials` switch now reaching keys that already exist, `validateIdForPath` on rotate and delete, `lastUsed` on the HTTP and MCP paths, the non-fatal list refresh, the clipboard guard, the switch semantics on both new toggles, and the two comment-formatting repairs. One extra fix came out of this: the generated key name is now capped so a long user name cannot produce a client ID that later fails path validation, which would have left the key unrotatable and unrevokable.

**Verification:** `npm run test:quick` exit 0 (48 suites, 611 tests, including 15 new checks), the integration suite 91 passed, the client bundle builds, and an end-to-end run against a four-worker cluster covering creation, use on the OpenAI-compatible API and the MCP gateway, the client-credentials exchange, rotation invalidating both credential types, expiry bounding the client-credentials path, both administrator switches applied without a restart, and 20 consecutive rotate-then-use sequences with no failures.

---

PR #2277, opened 2026-09-07
build(deps): bump tailwindcss to 4.3.3 and complete the v4 migration

manzke:
Supersedes intrafind/ihub-apps#2255 — it carries Dependabot's bump commit unchanged and adds the migration work that bump needs.

## Why #2255 can't merge on its own

Tailwind v4 is a major release. The version bump alone leaves the client unbuildable: every stylesheet fails with

```
[postcss] It looks like you're trying to use `tailwindcss` directly as a PostCSS plugin.
The PostCSS plugin has moved to a separate package...
```

That is what turned **Build Check** and the **axe-core** accessibility job red on #2255 (the a11y job runs `npm run dev`, which hits the same PostCSS error).

## What this adds

Ran the official `@tailwindcss/upgrade` codemods, then did by hand what they don't cover.

### Build setup

- `client/postcss.config.js` now uses `@tailwindcss/postcss`. `autoprefixer` is removed — v4 does vendor prefixing itself.
- `client/tailwind.config.js` is replaced by **`client/tailwind.css`**, holding the theme (`@theme`), the typography plugin (`@plugin`), the `data-theme="dark"` variant (`@custom-variant`) and the template scan paths (`@source`).

  The client has four CSS entry points — `src/App.css`, `office/office.css`, `extension/extension.css`, `nextcloud/nextcloud.css` — that all render components from `src/`. Each now starts with:

  ```css
  @import 'tailwindcss' source(none);
  @import '../tailwind.css';
  ```

  `source(none)` is deliberate: v4's automatic source detection is scoped to each entry stylesheet's own directory, which would miss `src/` and silently strip almost every utility from the Office, extension and Nextcloud bundles.

### Utility renames (codemods, 283 files)

| v3 | v4 |
| --- | --- |
| `flex-shrink-0`, `flex-grow` | `shrink-0`, `grow` |
| `shadow-sm`, `shadow` | `shadow-xs`, `shadow-sm` (same shift for `rounded`, `backdrop-blur`, `drop-shadow`) |
| `outline-none` | `outline-hidden` |
| `break-words` | `wrap-break-word` |
| `bg-gradient-to-*` | `bg-linear-to-*` |
| `!bg-gray-400` | `bg-gray-400!` |
| `z-[60]`, `max-w-[10rem]`, … | `z-60`, `max-w-40`, … |

### Done by hand

- **`bg-opacity-*` / `ring-opacity-*` (47 uses).** v4 removed these rather than renaming them, so the codemods leave them behind and they compile to nothing — every modal overlay would have rendered fully opaque. Converted to slash syntax: `bg-black bg-opacity-50` → `bg-black/50`.
- **Three codemod false positives reverted**, where the tool rewrote non-class text: an `outline` canvas action id (`FloatingToolbox.jsx`, `useCanvas.js`, `useCanvasEditing.js`, which would have broken the i18n key `canvas.outline` and the action's `switch` case) and the phrase `"shadow text"` in a `passageMatcher.js` comment.
- **v3 Preflight defaults restored** in `tailwind.css` so the upgrade is visually neutral: default border colour (`currentcolor` → `gray-200`), placeholder colour, button `cursor: pointer`, and the bare `ring` width/colour (1px `currentcolor` → 3px `blue-500`). Each block is annotated and can be dropped deliberately later.
- **Kept the v3 sRGB indigo ramp** — v4's default oklch indigo is noticeably more saturated and would shift every accent colour in the UI.

## Verification

- Diffed the **generated CSS against a v3 baseline build**. Every class present in v3 but absent in v4 has an accounted-for counterpart, and the declarations for the rescaled utilities match their v3 originals byte for byte (`.shadow-sm` v3 == `.shadow-xs` v4; `.rounded` v3 `.25rem` == `.rounded-sm` v4 `var(--radius-sm)` == `.25rem`). Confirmed the dark variant compiles to `:where([data-theme=dark], [data-theme=dark] *)` and that all four bundles contain the utilities used in `src/`.
- `npm run build` (client) — passes, all four entry bundles.
- `npm run test:unit` — 651 passed.
- `npm run test:a11y` (axe-core, WCAG 2.2 AA) — 3 passed, i.e. the job that was failing on #2255.
- `npm run lint:fix && npm run format:fix` — clean.
- Loaded the running app and checked the home, login and admin pages render correctly.

## Intentional upstream behaviour changes not compensated for

- `hover:` now only applies on devices that actually support hover (`@media (hover: hover)`).
- `space-x-*` / `space-y-*` use `> :not(:last-child)` instead of `> * + *`.

Both are deliberate upstream v4 changes with no v3 shim; flagging them since they're the likeliest source of any subtle difference.

https://claude.ai/code/session_0198CWwUnwLRWihHvSC6bAcR

---
_Generated by [Claude Code](https://claude.ai/code/session_0198CWwUnwLRWihHvSC6bAcR)_


manzke:
### Status on `33123dd` (rebased onto `main` @ `2ee64ad`)

Every functional check is green:

| Check | Result |
| --- | --- |
| Build Check | ✅ |
| test-quick | ✅ |
| test-integration | ✅ |
| Server boots | ✅ |
| axe-core | ✅ |
| auto-lint-format | ✅ |
| CodeQL + Analyze (js/ts, actions) | ✅ |

`mergeable_state` is `clean` — no conflict with `main`.

**One red check I'm standing down on: `copilot-pull-request-reviewer`.** It failed with an account-level quota error before reviewing anything, so nothing in this diff can affect it:

```
errorType: 'quota',
statusCode: 402,
[cause]: [Error: You have exceeded your monthly quota (Request ID: B442:28AC7B:9CF045:ADB85F:6A9ED33E)]
```

This is the GitHub Copilot code-review app's monthly quota for the org, triggered when the PR was marked ready for review. It's deterministic rather than flaky, so I haven't spent a re-run on it — it would return the same 402. It needs a quota top-up or the check made non-blocking; there's no fix available from inside this PR.

I also re-verified the rebase rather than trusting it: the diff against the new `main` touches `useMermaidRenderer.js` only for the `shadow-sm`→`shadow-xs` / `rounded`→`rounded-sm` / `backdrop-blur-sm`→`backdrop-blur-xs` renames, so #2276's mermaid hover fix is intact underneath, and #2276 introduced no v3-era classes that still need migrating. Client build and 654 unit tests pass locally on this head.

---

PR #2281, opened 2026-09-08
fix(auth): terminate OIDC provider session on logout (RP-Initiated Logout)

### LINKED ISSUE #2279: Logging out of iHub does not terminate the OIDC provider's SSO session
## Summary

**After internal discussion: this is not a bug.**

Logging out of iHub while authenticated via an OIDC provider (Keycloak,
verified end-to-end; Entra ID / Auth0 / ADFS by design, not yet verified
against a live tenant) only cleared iHub's own auth cookie. The provider's
own SSO session stayed active in the browser, so the next login attempt -
by the same user, or by a different person on a shared/kiosk device - was
silently re-authenticated by the provider without ever showing a login
form.

**Status: fixed** Branch [`fix/oidc-rp-initiated-logout`](https://github.com/intrafind/ihub-apps/pull/2281),
commit `a0ccdc0a`. This issue documents the finding for traceability; see the
PR for the fix.

## Impact

- Deviates from [OIDC RP-Initiated Logout 1.0](https://openid.net/specs/openid-connect-rpinitiated-1_0.html)
  and from session-termination expectations in NIST SP 800-63B ("Session
  tokens SHALL be erased or invalidated by the session subject when the
  subscriber logs out").
- Shared/kiosk devices: a user clicking "Log out" so a colleague can log
  in could result in the colleague being silently authenticated as the
  first user - no credential prompt, full account access transferred
  without anyone entering credentials.
- Users saw a "logged out" state that was misleading with respect to the
  IdP session.

## Current (pre-fix) behavior

- `server/middleware/oidcAuth.js`: the ID token returned by the provider
  was decoded once (to read a `groups` claim some IdPs only put there),
  then discarded - never persisted anywhere.
- `POST /api/auth/logout` (`server/routes/auth.js`) only cleared the
  local `authToken` cookie. No call to the provider's
  `end_session_endpoint`, no `id_token_hint`, no redirect to the IdP.
- `client/src/shared/contexts/AuthContext.jsx` (`logout()`) redirected
  the browser straight back to iHub's own home page, never to the
  provider.
- `oidcProviderSchema` (`server/validators/platformConfigSchema.js`) had
  no field for a logout/`end_session_endpoint` URL at all.

## Root cause

Investigated whether this was a deliberate design choice (e.g.
"local-only logout" to preserve a shared SSO session across other apps
on purpose) versus an oversight. Evidence points to oversight:

- `docs/ADFS-AUTHENTICATION-GUIDE.md` already documented a `logoutURL`
  provider field for exactly this purpose - it was never wired up in any
  code path.
- iHub's own OIDC **Identity Provider** side (`server/routes/oauth.js`,
  `/api/oauth/logout`) already implements RP-Initiated Logout correctly
  to spec - the pattern is understood and applied correctly elsewhere,
  just never mirrored on iHub's own relying-party side.
- No code comment, config toggle, or doc anywhere discussed a deliberate
  choice to keep OIDC logout local-only.

## Fix

Opt-in OIDC RP-Initiated Logout per provider, via a new optional
`endSessionURL` field (`server/validators/platformConfigSchema.js:23`).
No behavior change for any provider that leaves it unset - backward
compatible, no config migration needed.

1. **Login** (`server/middleware/oidcAuth.js`): the ID token is passed
   out of the Passport verify callback via `done(null, user, { idToken })`
   - Passport's `info` argument, not the user object (which can be
   persisted to `contents/config/users.json`; the ID token must never
   end up there). Stored in a new httpOnly `oidcLogoutHint` cookie, only
   when the provider has `endSessionURL` configured.
2. **Logout, step 1** (`POST /api/auth/logout`, `server/routes/auth.js:560`):
   returns `oidcLogoutRequired`, computed from both the *current*
   session's `authMode` (`req.user.authMode === 'oidc'`, from the
   verified JWT) and the hint cookie's presence - never reads its
   content here. Only clears the hint cookie itself when it won't be
   needed downstream.
3. **Logout, step 2** (`GET /api/auth/oidc-logout`, `server/routes/auth.js:862`):
   reached only via a real top-level browser navigation (never
   fetch/XHR). Reads the hint cookie, redirects to
   `{endSessionURL}?id_token_hint=...&post_logout_redirect_uri=...&client_id=...`.
   Any missing/malformed/unknown-provider/malformed-URL state falls back
   safely to local-only logout.
4. **Client** (`AuthContext.jsx`): `logout()` branches on
   `oidcLogoutRequired` to pick the landing target.
5. **Admin UI** (`OidcProvidersSection.jsx`): new "End Session URL" field
   next to Callback URL, with a live-computed hint showing the exact
   URL to register at the provider, and prefilled defaults for the
   Keycloak/Auth0/Microsoft templates.

**Security-conscious design:** the raw ID token only ever exists in the
httpOnly cookie and the `Location` header of the final redirect - never
in a JS-readable API response body (closes a CORS-misconfiguration /
browser-extension exposure path considered during design).

## Provider coverage

| Provider | End-session endpoint | Status |
| --- | --- | --- |
| Keycloak | `{issuer}/protocol/openid-connect/logout` | Verified end-to-end |
| Microsoft Entra ID | `.../oauth2/v2.0/logout` | Spec-compliant, untested live |
| Auth0 | `https://{domain}/oidc/logout` (not legacy `/v2/logout`) | Spec-compliant, untested live |
| ADFS | `.../adfs/oauth2/logout` | Spec-compliant, untested live (no practical local test env) |
| Google | no `end_session_endpoint` exists | Out of scope - needs a different, login-side mitigation |

## Bugs found and fixed during review (4 total)

1. Stale `oidcLogoutHint` cookie could redirect a non-OIDC logout (dual-auth
   deployments) through an unrelated provider - fixed by tying
   `oidcLogoutRequired` to the current session's actual `authMode`.
2. Unguarded `new URL(provider.endSessionURL)` could crash the logout
   route on a malformed saved value - wrapped in try/catch.
3. Admin UI's registration hint ignored subpath deployments
   (`window.location.origin` alone) - now uses `getBasePath()`.
4. **Found via live end-to-end testing, not by review:** an earlier fix
   for #1 had `POST /api/auth/logout` unconditionally clear the hint
   cookie, deleting it before the follow-up `GET /api/auth/oidc-logout`
   request ever got to read it - silently downgrading every OIDC logout
   back to local-only. Neither the security review nor the 8-angle code
   review caught this; only a real login-logout-login cycle against a
   live Keycloak did. Fixed by only clearing the cookie in step 1 when it
   won't be consumed by step 2.

## Testing

- `server/tests/oidc-logout.test.js`: 15 tests (redirect construction,
  fallback paths, cookie lifecycle across both requests, the
  X-Forwarded-Host dev-proxy fix, the bug-4 regression).
- Manual end-to-end against a local Keycloak (Docker): login, logout,
  re-login now requires a real credential prompt instead of silently
  reusing the provider's SSO session.
- Full design/root-cause writeup: `concepts/2026-09-07 OIDC RP-Initiated
  Logout Fix.md`.

## Out of scope / follow-ups

- Google-specific mitigation (`prompt=select_account` at login) -
  different mechanism, no `end_session_endpoint` exists.
- Live verification against real Entra ID / Auth0 / ADFS tenants.
- Two low-severity edge cases noted in the concept doc (client never
  following through to step 2; a narrow two-tab logout race) - not
  currently addressed or tested.

## References

- https://openid.net/specs/openid-connect-rpinitiated-1_0.html
- https://pages.nist.gov/800-63-4/sp800-63b.html

Labels: auth, backend, frontend, documentation, testing, i18n

jwalteri:
## Summary

Logging out of iHub while authenticated via an OIDC provider (Keycloak,
verified end-to-end; Entra ID / Auth0 / ADFS by design, not yet verified
against a live tenant) only cleared iHub's own auth cookie. The provider's
own SSO session stayed active in the browser, so the next login attempt -
by the same user, or by a different person on a shared/kiosk device - was
silently re-authenticated without ever showing a login form.

Implements opt-in OIDC RP-Initiated Logout per provider via a new
`endSessionURL` field. Fully backward compatible - no behavior change for
any provider that leaves it unset, no config migration needed.

Fixes #2279 

## What changed

- `server/validators/platformConfigSchema.js`: new optional `endSessionURL`
  field on `oidcProviderSchema`
- `server/middleware/oidcAuth.js`: the ID token is passed out of the OIDC
  verify callback via Passport's `info` argument (never persisted to the
  user object, JWT, or `users.json`) and stashed in a new httpOnly
  `oidcLogoutHint` cookie, only when the provider supports it
- `server/routes/auth.js`:
  - `POST /api/auth/logout` now signals `oidcLogoutRequired`, tied to the
    *current* session's `authMode` (not just cookie presence)
  - new `GET /api/auth/oidc-logout` builds the provider's
    `end_session_endpoint` redirect from the hint cookie, with a safe
    fallback to local-only logout on any missing/malformed state
- `client/src/shared/contexts/AuthContext.jsx`: `logout()` follows up with
  a top-level navigation to the new redirect route when required
- `client/.../OidcProvidersSection.jsx`: new **End Session URL** admin
  field with a live registration hint, prefilled defaults for
  Keycloak/Auth0/Microsoft templates
- `server/tests/oidc-logout.test.js`: 15 new tests
- Docs: `docs/oidc-authentication.md`, `docs/ADFS-AUTHENTICATION-GUIDE.md`,
  `docs/releases/5.5.0/fixes.md`
- Full root-cause analysis and design rationale:
  `concepts/2026-09-07 OIDC RP-Initiated Logout Fix.md`

## Security-conscious design

The raw ID token only ever exists in the httpOnly cookie and the
`Location` header of the final redirect - never in a JS-readable API
response body (an earlier draft returned it via JSON; reworked into a
two-step redirect specifically to close that exposure path).

## Review process

Both a dedicated security review and an 8-angle code review were run
against this diff before opening this PR. No HIGH/MEDIUM security
findings survived. The code review found and fixed 3 real bugs; live
end-to-end testing against a real Keycloak instance then caught a 4th,
more serious regression that neither review had caught (an earlier fix
for bug 1 cleared the hint cookie one request too early, silently
downgrading every OIDC logout back to local-only). All four are detailed,
with root cause and fix, in the concept doc linked above.

A CodeQL "missing CSRF middleware" alert on `server/tests/oidc-logout.test.js`
has been reviewed and dismissed as "Used in tests" - the test file
instantiates the real production route handlers directly but is never
deployed or reachable; the underlying (pre-existing, unrelated to this PR)
lack of CSRF protection on `/api/auth/logout` was already assessed during
the security review as low-impact (self-targeting forced logout only, no
data/account exposure), consistent with every other auth route in this
app.

## Testing

- `server/tests/oidc-logout.test.js`: 15/15 passing
- Manual end-to-end against a local Keycloak (Docker): login, logout,
  re-login now requires a real credential prompt instead of silently
  reusing the provider's SSO session
- `npm run lint:fix && npm run format:fix`: clean

## Out of scope / follow-ups

- Google-specific mitigation (no `end_session_endpoint` exists there)
- Live verification against real Entra ID / Auth0 / ADFS tenants
- Two low-severity edge cases noted in the concept doc (client never
  following through to the redirect step; a narrow two-tab logout race)

### DIFF
--- a/client/src/features/admin/components/platform-form/OidcProvidersSection.jsx
+++ b/client/src/features/admin/components/platform-form/OidcProvidersSection.jsx
@@ -2,6 +2,7 @@ import { useState } from 'react';
 import Icon from '../../../../shared/components/Icon';
 import { CredentialRefSelect } from '../OpenApiToolEditor';
 import GroupMultiSelect from '../GroupMultiSelect';
+import { getBasePath } from '../../../../utils/runtimeBasePath';
 
 // OIDC Provider Templates
 const OIDC_PROVIDER_TEMPLATES = {
@@ -11,6 +12,9 @@ const OIDC_PROVIDER_TEMPLATES = {
     authorizationURL: 'https://${AUTH0_DOMAIN}/authorize',
     tokenURL: 'https://${AUTH0_DOMAIN}/oauth/token',
     userInfoURL: 'https://${AUTH0_DOMAIN}/userinfo',
+    // The OIDC-compliant endpoint - not the legacy /v2/logout, which uses
+    // different (non-standard) query parameters.
+    endSessionURL: 'https://${AUTH0_DOMAIN}/oidc/logout',
     scope: ['openid', 'profile', 'email'],
     groupsAttribute: 'groups',
     pkce: true,
@@ -22,6 +26,8 @@ const OIDC_PROVIDER_TEMPLATES = {
     authorizationURL: 'https://accounts.google.com/o/oauth2/v2/auth',
     tokenURL: 'https://www.googleapis.com/oauth2/v4/token',
     userInfoURL: 'https://www.googleapis.com/oauth2/v2/userinfo',
+    // No endSessionURL: Google has no end_session_endpoint, so RP-Initiated
+    // Logout isn't possible here - left unset deliberately.
     scope: ['openid', 'profile', 'email'],
     groupsAttribute: 'groups',
     pkce: true,
@@ -33,6 +39,7 @@ const OIDC_PROVIDER_TEMPLATES = {
     authorizationURL: 'https://login.microsoftonline.com/common/oauth2/v2.0/authorize',
     tokenURL: 'https://login.microsoftonline.com/common/oauth2/v2.0/token',
     userInfoURL: 'https://graph.microsoft.com/v1.0/me',
+    endSessionURL: 'https://login.microsoftonline.com/common/oauth2/v2.0/logout',
     scope: ['openid', 'profile', 'email', 'User.Read'],
     groupsAttribute: 'groups',
     pkce: true,
@@ -46,6 +53,8 @@ const OIDC_PROVIDER_TEMPLATES = {
     tokenURL: 'https://${KEYCLOAK_SERVER}/realms/${KEYCLOAK_REALM}/protocol/openid-connect/token',
     userInfoURL:
       'https://${KEYCLOAK_SERVER}/realms/${KEYCLOAK_REALM}/protocol/openid-connect/userinfo',
+    endSessionURL:
+      'https://${KEYCLOAK_SERVER}/realms/${KEYCLOAK_REALM}/protocol/openid-connect/logout',
     scope: ['openid', 'profile', 'email'],
     groupsAttribute: 'groups',
     pkce: true,
@@ -315,6 +324,42 @@ function OidcProvidersSection({ config, onChange, t, availableGroups = [] }) {
                     <p className="text-xs text-blue-600 dark:text-blue-400 mt-1">
                       <Icon name="information-circle" className="h-3 w-3 inline mr-1" />
                       Auto-generated: /api/auth/oidc/{provider.name}/callback
+                    </p>
+                  )}
+                </div>
+                <div className="md:col-span-2">
+                  <label className="block text-sm font-medium text-gray-700 dark:text-gray-300 mb-1">
+                    {t('admin.auth.endSessionURL', 'End Session URL')}{' '}
+                    <span className="text-xs text-gray-500 dark:text-gray-400 font-normal">
+                      {t(
+                        'admin.auth.endSessionURLOptional',
+                        '(Optional - enables full logout at the provider)'
+                      )}
+                    </span>
+                  </label>
+                  <input
+                    type="url"
+                    placeholder={t(
+                      'admin.auth.endSessionURLPlaceholder',
+                      'https://your-provider/.../logout'
+                    )}
+                    value={provider.endSessionURL || ''}
+                    onChange={e => updateOidcProvider(index, 'endSessionURL', e.target.value)}
+                    className="w-full px-3 py-2 border border-gray-300 rounded-md shadow-sm focus:ring-blue-500 focus:border-blue-500 sm:text-sm"
+                  />
+                  <p className="text-xs text-gray-500 dark:text-gray-400 mt-1">
+                    {t(
+                      'admin.auth.endSessionURLHelp',
+                      'When set, logging out of iHub also ends the session at the provider (RP-Initiated Logout), so the next login always shows a real login prompt instead of silently reusing an active provider session. Leave empty to log out of iHub only.'
+                    )}
+                  </p>
+                  {provider.endSessionURL && (
+                    <p className="text-xs text-blue-600 dark:text-blue-400 mt-1">
+                      <Icon name="information-circle" className="h-3 w-3 inline mr-1" />
+                      {t(
+                        'admin.auth.endSessionURLRegisterHint',
+                        'Register {{url}} as a valid post-logout redirect URI at your provider, or logout will fail with an error there.'
+                      ).replace('{{url}}', `${window.location.origin}${getBasePath()}/*`)}
                     </p>
                   )}
                 </div>
--- a/server/middleware/oidcAuth.js
+++ b/server/middleware/oidcAuth.js
@@ -229,7 +229,13 @@ export function configureOidcProviders() {
               sessionId
             );
 
-            return done(null, validatedUser);
+            // Pass the raw ID token out via Passport's `info` argument rather than
+            // attaching it to validatedUser: that object flows into
+            // validateAndPersistExternalUser() above, which can persist user fields
+            // to contents/config/users.json, and the ID token must never end up
+            // there. `info` is only used transiently by createOidcCallbackHandler
+            // to set the (also transient, httpOnly) oidcLogoutHint cookie.
+            return done(null, validatedUser, { idToken: params?.id_token });
           } catch (error) {
             authDebugService.log(
               'oidc',
@@ -266,6 +272,24 @@ export function configureOidcProviders() {
         component: 'OidcAuth',
         providerName: provider.name
       });
+
+      // endSessionURL enables RP-Initiated Logout (GET /api/auth/oidc-logout), but the
+      // provider must separately allow-list iHub's own URL as a post-logout redirect
+      // target - a setup step iHub cannot verify from here. Surface it now so it's
+      // caught during setup, not by a user hitting the IdP's error page at logout.
+      if (provider.endSessionURL) {
+        logger.warn(
+          'OIDC provider has endSessionURL configured - your iHub URL must be allow-listed ' +
+            'at the provider as a valid post-logout redirect target (e.g. Keycloak: client ' +
+            '"Valid post logout redirect URIs"), otherwise logout will fail with an IdP-side ' +
+            'error. See docs/oidc-authentication.md.',
+          {
+            component: 'OidcAuth',
+            providerName: provider.name,
+            endSessionURL: provider.endSessionURL
+          }
+        );
+      }
     } catch (error) {
       logger.error('Failed to configure OIDC provider', {
         component: 'OidcAuth',
@@ -735,6 +759,20 @@ export function createOidcCallbackHandler(providerName) {
             persistedUser: user.persistedUser || false
           }
         });
+
+        // Carry the ID token for a possible later RP-Initiated Logout
+        // (https://openid.net/specs/openid-connect-rpinitiated-1_0.html) in a
+        // dedicated httpOnly cookie. Only set when the provider actually supports
+        // it (endSessionURL configured) - no cookie, no behavior change otherwise.
+        // Consumed exclusively by GET /api/auth/oidc-logout; never exposed to
+        // client JS or included in any JSON API response.
+        if (provider.endSessionURL && info?.idToken) {
+          res.cookie(
+            'oidcLogoutHint',
+            JSON.stringify({ provider: providerName, idToken: info.idToken }),
+            getAuthCookieOptions(expiresIn * 1000, req)
+          );
+        }
 
         authDebugService.log(
           'oidc',
--- a/server/routes/auth.js
+++ b/server/routes/auth.js
@@ -3,7 +3,8 @@ import { createAuthorizationMiddleware } from '../utils/authorization.js';
 import {
   getConfiguredProviders,
   createOidcAuthHandler,
-  createOidcCallbackHandler
+  createOidcCallbackHandler,
+  configuredProviders
 } from '../middleware/oidcAuth.js';
 import { loginLdapUser, getConfiguredLdapProviders } from '../middleware/ldapAuth.js';
 import { processNtlmLogin, getNtlmConfig } from '../middleware/ntlmAuth.js';
@@ -19,6 +20,7 @@ import { sendBadRequest, sendAuthRequired, sendErrorResponse } from '../utils/re
 import { recordAuthEvent } from '../telemetry/metrics.js';
 import { logAudit } from '../services/AuditLogService.js';
 import { getAuthCookieOptions, getClearAuthCookieOptions } from '../utils/cookieSettings.js';
+import { buildPublicBaseUrl } from '../utils/publicBaseUrl.js';
 
 /**
  * Sanitize and validate authentication input
@@ -542,8 +544,34 @@ export default function registerAuthRoutes(app) {
    * Logout (clear cookies and track logout)
    */
   app.post(buildServerPath('/api/auth/logout'), (req, res) => {
+    // Presence-only check: whether an OIDC RP-Initiated Logout redirect is
+    // needed. Deliberately does not read/parse the cookie's content here - the
+    // ID token it carries is only ever touched by GET /api/auth/oidc-logout,
+    // never by this JSON-responding endpoint.
+    //
+    // Also requires the *current* session to actually be OIDC-authenticated
+    // (req.user.authMode, decoded from the current authToken JWT - see
+    // jwtAuth.js). The cookie alone isn't enough: on a dual-auth deployment, a
+    // user could log in via OIDC (setting oidcLogoutHint), then log in again
+    // via local/LDAP/NTLM/Teams without an intervening logout - authToken gets
+    // overwritten but nothing else clears oidcLogoutHint, so without this
+    // check a later logout from that non-OIDC session would still redirect
+    // through the stale, unrelated OIDC provider's end_session_endpoint.
+    const oidcLogoutRequired =
+      req.user?.authMode === 'oidc' && req.cookies?.oidcLogoutHint !== undefined;
+
     // Clear the authentication cookie
     res.clearCookie('authToken', getClearAuthCookieOptions(req));
+    // Only clear the hint here when it WON'T be used - i.e. when this session
+    // isn't OIDC, so the client will redirect straight to /?logout=true and
+    // never call GET /api/auth/oidc-logout at all. When oidcLogoutRequired is
+    // true, that route is the one responsible for reading *and* clearing this
+    // cookie (see below) - clearing it here too would delete it before the
+    // client's follow-up navigation ever gets to read it, silently downgrading
+    // every OIDC logout to a local-only one.
+    if (!oidcLogoutRequired) {
+      res.clearCookie('oidcLogoutHint', getClearAuthCookieOptions(req));
+    }
     recordAuthEvent(req.user?.authMode || 'unknown', 'logout');
     if (req.user && req.user.id !== 'anonymous') {
       logAudit({
@@ -579,7 +607,8 @@ export default function registerAuthRoutes(app) {
 
     res.json({
       success: true,
-      message: 'Logged out successfully'
+      message: 'Logged out successfully',
+      oidcLogoutRequired
     });
   });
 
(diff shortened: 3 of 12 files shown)


jwalteri:
This PR does not need to be merged, as the current behavior is considered
intended and not a bug. Offering OIDC RP-Initiated Logout as an optional
feature could still be considered separately.

server/middleware/oidcAuth.js
```diff
@@ -736,6 +760,20 @@ export function createOidcCallbackHandler(providerName) {
...
 
+        // Carry the ID token for a possible later RP-Initiated Logout
+        // (https://openid.net/specs/openid-connect-rpinitiated-1_0.html) in a
+        // dedicated httpOnly cookie. Only set when the provider actually supports
+        // it (endSessionURL configured) - no cookie, no behavior change otherwise.
+        // Consumed exclusively by GET /api/auth/oidc-logout; never exposed to
+        // client JS or included in any JSON API response.
+        if (provider.endSessionURL && info?.idToken) {
+          res.cookie(
+            'oidcLogoutHint',
```
manzke, line 771: This only ever **sets** the cookie, never clears it — which leaves a gap the `authMode` check in `POST /api/auth/logout` doesn't cover.

Log in via provider A (has `endSessionURL`), then via provider B (no `endSessionURL`) without an intervening logout: `authMode` is still `'oidc'` and the cookie still holds A's name plus A's stale ID token. Logout redirects to A's `end_session_endpoint`, and B's session is never terminated.

Making this an unconditional set-or-clear closes it, and makes cookie presence authoritative on its own (see my comment on `auth.js:560`):

```js
if (provider.endSessionURL && info?.idToken) {
  res.cookie('oidcLogoutHint', ..., getAuthCookieOptions(expiresIn * 1000, req));
} else {
  res.clearCookie('oidcLogoutHint', getClearAuthCookieOptions(req));
}
```

Worth also clearing it in the local/LDAP/NTLM/Teams login handlers, so a stale ID token stops riding along on every request after an auth-mode switch.

Three more things about this cookie:

**Size.** `JSON.stringify({provider, idToken})`, URL-encoded (every `"` becomes `%22`), against a ~4096-byte per-cookie limit. Entra ID tokens with a `groups` claim get close on their own. An oversized cookie is dropped by the browser **silently**, and the flow then degrades to local-only logout with no signal at all. Measuring the encoded value and warning past a threshold is cheap. Better still: store the provider name alone in that case and omit `id_token_hint` — the spec only *recommends* it, and `client_id` + `post_logout_redirect_uri` still terminates the session (some providers then show a confirmation page).

**Path.** `getAuthCookieOptions` sets no `path`, so this defaults to `/` and rides on every request to the origin — static assets, API, streaming — for the full 8h JWT lifetime, alongside the `authToken` JWT. That can exceed nginx's default `large_client_header_buffers 4 8k`. Only two handlers read it, both under `/api/auth`, so `path: buildServerPath('/api/auth')` (matched on every clear) keeps it off everything else.

**It can reach the logs.** `oidcAuth.js:701` already logs `cookies: req.headers.cookie` verbatim on a state-verification failure, and `exactSensitiveFields` in `server/utils/logger.js` has no `cookie`/`cookies` entry — so the raw ID token can land in logs. The PR description's claim that it never exists outside the httpOnly cookie and the `Location` header isn't quite right today. Logging cookie *names* there, plus adding `cookie`/`cookies` to the redaction list as a backstop, fixes it.

---

server/routes/auth.js
```diff
@@ -542,8 +544,34 @@ export default function registerAuthRoutes(app) {
...
+    // Also requires the *current* session to actually be OIDC-authenticated
+    // (req.user.authMode, decoded from the current authToken JWT - see
+    // jwtAuth.js). The cookie alone isn't enough: on a dual-auth deployment, a
+    // user could log in via OIDC (setting oidcLogoutHint), then log in again
+    // via local/LDAP/NTLM/Teams without an intervening logout - authToken gets
+    // overwritten but nothing else clears oidcLogoutHint, so without this
+    // check a later logout from that non-OIDC session would still redirect
+    // through the stale, unrelated OIDC provider's end_session_endpoint.
+    const oidcLogoutRequired =
+      req.user?.authMode === 'oidc' && req.cookies?.oidcLogoutHint !== undefined;
```
manzke, line 561: Gating on `authMode === 'oidc'` ties the feature to a live iHub JWT, and that re-opens the shared-device case the PR exists to close.

The provider's SSO session routinely outlives iHub's JWT (default 8h). Once the JWT expires, `/auth/status` answers anonymous → `state.isAuthenticated` is false → `AuthContext.jsx:503` skips the POST entirely → local-only logout. The user is then offered "Log in", and clicking it silently re-authenticates them as… themselves, or on a kiosk, as whoever used the machine before. Exactly the scenario in the concept doc's Impact section.

The `authMode` gate exists to defend against a stale hint surviving a non-OIDC login. Clearing the hint at those login sites instead (see my comment on `oidcAuth.js:771`) makes the cookie authoritative by construction, so this can become presence-only and keeps working past JWT expiry:

```js
const oidcLogoutRequired = req.cookies?.oidcLogoutHint !== undefined;
```

`AuthContext.logout()` then also needs to drop its `if (state.isAuthenticated)` guard — the endpoint is idempotent, and it's the only thing that can tell the client whether a hint is still present.

**Still not fully covered either way:** after JWT expiry there's no "Log out" button to press at all. Closing that needs a login-side `prompt=login` / `max_age=0` option on the authorization request — a separate feature, and also the only possible mitigation for Google, which the concept doc already lists as out of scope.

---

server/routes/auth.js
```diff
@@ -813,6 +842,121 @@ export default function registerAuthRoutes(app) {
...
+    let hint;
+    let parseError = false;
+    try {
+      hint = req.cookies?.oidcLogoutHint ? JSON.parse(req.cookies.oidcLogoutHint) : null;
+    } catch {
+      hint = null;
+      parseError = true;
+    }
+
+    res.clearCookie('oidcLogoutHint', getClearAuthCookieOptions(req));
```
manzke, line 874: **This unauthenticated GET is reachable cross-site, and that's a new surface.**

The hint cookie is `SameSite=Lax`, and Lax cookies *are* sent on cross-site top-level navigations. So any third-party page can `window.open('https://ihub.example.com/api/auth/oidc-logout')` and:

1. terminate the victim's SSO session at **every application federated to that IdP** — not self-targeting like the pre-existing `/api/auth/logout` gap the PR description assesses; and
2. consume the hint cookie, so the rest of that iHub session's logout silently degrades to local-only — i.e. disarm the feature this PR adds.

`sameSite: 'strict'` on this one cookie closes both. It's only ever read from a same-site navigation that iHub's own page initiates, and the cross-site *set* at the OIDC callback still works — SameSite governs when a cookie is sent, not whether it may be stored.

Separately: this route clears the hint but **not** `authToken`. Hit directly while signed in, it ends the provider session and leaves iHub logged in. Clearing `authToken` here too makes the endpoint self-consistent — but only once a hint has actually been presented. An unconditional clear would turn this into a cross-site forced logout for iHub itself, which isn't possible today (a cross-site form POST to `/api/auth/logout` wouldn't carry the Lax `authToken`). With Strict on the hint, "hint present" implies "same-site", so the conditional is safe.

---

server/routes/auth.js
```diff
@@ -813,6 +842,121 @@ export default function registerAuthRoutes(app) {
...
+      logoutUrl = new URL(provider.endSessionURL);
+    } catch {
+      logger.warn('OIDC provider endSessionURL is not a valid URL - falling back', {
+        component: 'Auth',
+        providerName: provider.name
+      });
+      return res.redirect(fallbackUrl);
+    }
+    logoutUrl.searchParams.set('id_token_hint', hint.idToken);
+    logoutUrl.searchParams.set('post_logout_redirect_uri', `${baseUrl}/?logout=true`);
```
manzke, line 943: Two things here.

**`new URL()` accepts more than you want.** `javascript:` parses fine, and — more likely in practice — an unresolved `${VAR}` placeholder does too: `configCache.resolveEnvVars` keeps the placeholder verbatim and only warns when the env var is missing, and `${KEYCLOAK_SERVER}` is a perfectly legal URL host (it percent-encodes to `$%7Bkeycloak_server%7D`). So a missing `KEYCLOAK_SERVER` doesn't hit the `catch` — it sends the browser to a nonexistent host instead of taking the graceful fallback every other invalid-state branch takes. Two extra guards:

```js
if (logoutUrl.protocol !== 'https:' && logoutUrl.protocol !== 'http:') { /* fall back */ }
if (provider.endSessionURL.includes('${')) { /* fall back */ }
```

**`post_logout_redirect_uri` should be configurable.** It's hard-coded to `${baseUrl}/?logout=true`, which causes two problems:

- Providers that compare post-logout URIs exactly, or reject a query string in them, may refuse it. Entra ID is the one to check — the PR already flags it as unverified. And you can't just drop the query: `?logout=true` is load-bearing, it's what suppresses `autoRedirect` in `AuthContext.jsx:188` and `auth-gate.js:121`. Without it an `autoRedirect` deployment bounces straight back into the IdP.
- `baseUrl` comes from `X-Forwarded-Host` per request, so a deployment reachable under several hostnames has to register every one of them.

An optional per-provider `postLogoutRedirectURL` lets admins pin whatever their IdP will actually accept, keeping `?logout=true` on it.

---

server/tests/oidc-logout.test.js
```diff
@@ -0,0 +1,238 @@
...
+ * OIDC RP-Initiated Logout — unit tests.
+ *
+ * Locks in:
+ * - POST /api/auth/logout signals oidcLogoutRequired only when the *current*
+ *   session's authMode is 'oidc' AND the hint cookie is present - not from
+ *   cookie presence alone (regression: a stale hint cookie surviving a later
+ *   non-OIDC login must not redirect that session's logout through an
+ *   unrelated provider).
+ * - It never reads/uses the ID token the cookie carries.
+ * - It always clears both authToken and oidcLogoutHint, regardless of mode.
```
manzke, line 11: **These 15 tests don't run in CI.** All checks green isn't evidence for them.

The root jest config's `testMatch` is `tests/integration/**`, `tests/unit/server/**`, `tests/unit/client/**` — `server/tests/oidc-logout.test.js` matches none of them, and no npm script references the file. CI runs only `test:quick` and `test:integration:ci`; `cd server && npm test` (whose bare jest config *would* pick it up) is in no workflow. There are ~70 other jest-style files under `server/tests/` in the same state, so this is pre-existing rot rather than something this PR introduced — but the new coverage is worth having actually execute.

Relocating to `tests/unit/server/` doesn't work here, incidentally: that config transforms `.js` to CJS, and `routes/auth.js` reaches `middleware/localAuth.js`, which uses `import.meta.url`. So the fix is CI wiring — a script running this file under the server's own native-ESM jest, chained into `test:quick`.

Small thing in this header: line 11 says "It always clears both authToken and oidcLogoutHint, regardless of mode", which the two tests below it (`does NOT clear the oidcLogoutHint cookie here when it WILL be used`) contradict.

Two invariants worth locking in that aren't covered:
- The hint cookie is set **only** when `endSessionURL` is present (the middleware side is untested).
- The ID token never reaches `validatedUser` / `users.json` — the whole reason the `info` argument is used instead of the user object. That's the design decision most likely to be undone by a future refactor with nothing failing.

---

client/src/features/admin/components/platform-form/OidcProvidersSection.jsx
```diff
@@ -318,6 +327,42 @@ function OidcProvidersSection({ config, onChange, t, availableGroups = [] }) {
...
+                      'When set, logging out of iHub also ends the session at the provider (RP-Initiated Logout), so the next login always shows a real login prompt instead of silently reusing an active provider session. Leave empty to log out of iHub only.'
+                    )}
+                  </p>
+                  {provider.endSessionURL && (
+                    <p className="text-xs text-blue-600 dark:text-blue-400 mt-1">
+                      <Icon name="information-circle" className="h-3 w-3 inline mr-1" />
+                      {t(
+                        'admin.auth.endSessionURLRegisterHint',
+                        'Register {{url}} as a valid post-logout redirect URI at your provider, or logout will fail with an error there.'
+                      ).replace('{{url}}', `${window.location.origin}${getBasePath()}/*`)}
```
manzke, line 362: `origin + basePath + '/*'` is Keycloak wildcard syntax. Providers that match post-logout URIs exactly (Entra ID, Auth0's Allowed Logout URLs) need the value character-for-character, which is `origin + basePath + '/?logout=true'` — what the server actually sends.

Showing the exact URI and mentioning the wildcard as the Keycloak-friendly alternative covers both, and gives the admin something they can copy straight into the allow-list either way.

---

shared/i18n/en.json
```diff
@@ -682,6 +682,11 @@
       "callbackURLOptional": "(Optional - Auto-generated if empty)",
       "callbackURLPlaceholder": "/api/auth/oidc/{{providerName}}/callback",
       "callbackURLHelp": "Leave empty to auto-generate. Override for custom domains or subpath deployments.",
+      "endSessionURL": "End Session URL",
+      "endSessionURLOptional": "(Optional - enables full logout at the provider)",
+      "endSessionURLPlaceholder": "https://auth.example.com/oidc/logout",
+      "endSessionURLHelp": "Also ends the session at the provider on logout, so the next login can't silently reuse it. Leave empty for local logout only.",
```
manzke, line 688: These two have drifted from the inline `t()` fallbacks in `OidcProvidersSection.jsx`:

- help: `Also ends the session at the provider on logout…` here vs. `When set, logging out of iHub also ends the session at the provider (RP-Initiated Logout), so the next login always shows a real login prompt…` in the JSX
- placeholder: `https://auth.example.com/oidc/logout` vs. `https://your-provider/.../logout`

Since the key exists, the JSON wins and the longer, more informative fallback text is dead. Worth picking one — the JSX wording explains *why* better. Same in `de.json`.

---

server/middleware/oidcAuth.js
```diff
@@ -266,6 +272,24 @@ export function configureOidcProviders() {
         component: 'OidcAuth',
         providerName: provider.name
       });
+
+      // endSessionURL enables RP-Initiated Logout (GET /api/auth/oidc-logout), but the
+      // provider must separately allow-list iHub's own URL as a post-logout redirect
+      // target - a setup step iHub cannot verify from here. Surface it now so it's
+      // caught during setup, not by a user hitting the IdP's error page at logout.
+      if (provider.endSessionURL) {
+        logger.warn(
```
manzke, line 281: This warns on every boot for a *correctly* configured provider. Surfacing the allow-list requirement at setup is the right instinct, but `logger.info` fits an unavoidable, permanent notice better — a warning that can never be resolved teaches operators to filter the log.

---

manzke (commented):
Reviewed against the merge base (`07f13e1`, 12 files / +889 −7 — the GitHub file list is inflated by a stale base). The design is sound and unusually well documented; the concept doc's account of the four bugs found along the way is honest and useful.

Everything below is implemented on **`claude/pr-2281-review-6u3fc7`**, branched from this PR's head, if you'd rather take the patch than the prose. One deliberate change there: the field is named **`logoutURL`**, not `endSessionURL` — that's the name `docs/ADFS-AUTHENTICATION-GUIDE.md` already documented, so an admin who followed that guide gets a working logout instead of a silently dead config key, and the doc needs no correction.

## Not visible in this diff: the endpoint is rate limited

`READ_ONLY_AUTH_PATHS` in `server/middleware/rateLimiting.js` exempts `/logout` from the strict credential limiter, and its wildcard is `normalized.startsWith('/oidc/')` — which does **not** match `/oidc-logout` (no slash). So `GET /api/auth/oidc-logout` runs under `authApiLimiter`: 30 requests / 15 min in the shipped `platform.json`, and behind two proxy hops every caller shares one counter (that file's own comment explains this at length).

A 429 there lands *after* `authToken` has already been cleared: the user is logged out of iHub, still signed in at the provider, looking at a JSON error — the precise failure this PR exists to prevent, now reachable through a rate limit. The comment directly above the set says `/api/auth/logout` "must always be able to clear a session"; the second half of that flow was missed. Verified live: 41 consecutive requests to the endpoint all 429'd before the fix and all 302 after.

## Inline

Seven comments below, roughly in severity order: the `authMode` gate re-opening the shared-device case after JWT expiry; the hint cookie never being cleared (so it can point at the wrong provider); cross-site reachability of the new GET; ID-token cookie size, path scoping, and an existing log path that captures it; a non-configurable `post_logout_redirect_uri`; `new URL()` accepting `javascript:` and unresolved `${VAR}`; and the tests not running in CI.

## Questions not asked

- **Should iHub logout revoke the tokens iHub itself issued?** iHub is an OP as well as an RP (`/api/oauth/logout`, advertised in `.well-known`). Today "log out" leaves downstream clients' access/refresh tokens (Office add-in, Nextcloud, extension) fully alive. Same threat model, opposite direction — out of scope here, but it means "logged out" still isn't complete.
- **Provider-initiated logout.** Signing out at the IdP or in another app leaves the iHub session running for the full JWT lifetime. Front-channel/back-channel logout isn't in the follow-up list.
- No `state` on the outbound end_session request, while iHub's own OP-side implementation validates one. Spec-optional; noting the asymmetry.

## On the framing

Your [comment](https://github.com/intrafind/ihub-apps/pull/2281#issuecomment-5581425865) says the current behavior is intended and that RP-Initiated Logout could be considered separately as an optional feature — but this PR **is** already opt-in and backward compatible; nothing changes for a provider that leaves the field unset. The conflict is in the framing, not the code: `Fixes #2279`, the changelog entry under `fixes.md`, and the concept doc's "root cause classification: oversight" all assert bug-ness.

Moving the entry to `docs/releases/5.5.0/features.md`, changing `Fixes #2279` to `Refs #2279`, and rewriting the root-cause section as "why this was never implemented" rather than "why it's a bug" puts it exactly where you said it could be considered — with no code change at all. Done that way on the review branch.

---

PR #2337, opened 2026-09-11
fix(admin): clear axios default Content-Type for FormData uploads

### LINKED ISSUE #2334: Admin backup/import (and other admin FormData uploads) fail with HTTP 400 "No ZIP file uploaded" — axios instance default Content-Type not cleared for FormData
## Summary

Any admin request that sends a `FormData` body through `makeAdminApiCall` (e.g. `POST /api/admin/backup/import`, `POST /api/admin/ui/upload-asset`) fails with `HTTP 400`, and on the backup-import endpoint specifically with `"No ZIP file uploaded"` — even though the selected file is valid and the client-side code looks correct.

Observed on a live instance (`ghcr.io/intrafind/ihub-apps:5.4.14`): server logs show only

```
{"component":"AdminBackup","level":"info","message":"Starting configuration import"}
```

and then nothing else — no "Extracting ZIP file", no error log — because `multer`'s `upload.single('backup')` never sees a `multipart/form-data` request and leaves `req.file` undefined, so `importConfig` hits:

```js
// server/routes/admin/backup.js
if (!req.file) {
  return sendBadRequest(res, 'No ZIP file uploaded');
}
```

## Root cause

`client/src/api/client.js` creates the shared axios instance with an **instance-level default header**:

```js
const apiClient = axios.create({
  baseURL: API_URL,
  headers: { 'Content-Type': 'application/json' },
  ...
});
```

`client/src/api/adminApi.js` (`makeAdminApiCall`) tries to strip `Content-Type` for `FormData` bodies, but only on the **per-request** headers object, which is normally empty:

```js
axiosConfig.data = body;
if (isFormData) {
  Object.keys(axiosConfig.headers).forEach(headerKey => {
    if (headerKey.toLowerCase() === 'content-type') delete axiosConfig.headers[headerKey];
  });
}
```

Since `axiosConfig.headers` has nothing to delete, the instance-level default `Content-Type: application/json` from `client.js` still applies. Axios's default `transformRequest` sees an explicit JSON content type together with a `FormData` payload and serializes the `FormData` to JSON (`formToJSON`) instead of sending it as `multipart/form-data`. The ZIP file itself is therefore never transmitted — the server correctly (but unhelpfully) reports "no file uploaded".

This looks like a regression from #1823 ("Unify makeAdminApiCall payload handling on body and remove fetch/axios split", merged 2026-07-06), which removed a separate `fetch()` path that previously handled `FormData` uploads without this Content-Type conflict.

## Steps to reproduce

1. Run iHub (reproduced on `5.4.14`, current `main` has the same code path).
2. 2. Admin → System → Configuration Backup & Import.
3. 3. Export a configuration backup (works fine — GET request, unaffected).
4. 4. Re-import the exported ZIP.
5. 5. Import fails immediately with `HTTP 400` / "No ZIP file uploaded", no corresponding server-side error log beyond "Starting configuration import".
## Suggested fix

Explicitly unset the header instead of deleting a (likely empty) per-request object, e.g. in `makeAdminApiCall`:

```js
if (isFormData) {
  axiosConfig.headers = { ...axiosConfig.headers, 'Content-Type': undefined };
}
```

`undefined` tells axios to omit the header entirely so it (or the browser) can set the correct `multipart/form-data; boundary=...` value, regardless of what the axios instance's default headers say.

## Scope

Affects every FormData upload routed through `makeAdminApiCall`, not just backup import — e.g. `POST /admin/ui/upload-asset` in `adminApi.js` uses the same helper.

ifjoergi:
Fixes #2334.

`makeAdminApiCall` only deleted `Content-Type` from the per-request headers object, which is normally empty. That left `apiClient`'s axios-instance default (`Content-Type: application/json`, set in `client.js`) in place, so axios's default `transformRequest` JSON-serialized `FormData` bodies instead of sending them as `multipart/form-data` — breaking backup import and any other admin `FormData` upload.

Explicitly set the header to `undefined` so it overrides the instance default regardless of what the per-request headers object contains.

### DIFF
--- a/client/src/api/adminApi.js
+++ b/client/src/api/adminApi.js
@@ -56,6 +56,12 @@ export const makeAdminApiCall = async (url, options = {}) => {
           delete axiosConfig.headers[headerKey];
         }
       });
+      // Deleting the key above only clears this per-request headers object, not
+      // apiClient's axios-instance default (Content-Type: application/json, set in
+      // client.js) — axios still merges that in and JSON-serializes the FormData
+      // instead of sending it as multipart. Set it to undefined to override the
+      // instance default and let axios/browser set the multipart boundary.
+      axiosConfig.headers['Content-Type'] = undefined;
     } else {
       axiosConfig.headers = {
         'Content-Type': 'application/json',


manzke:
fixed in #2336

ifjoergi:
Confirmed — #2336 applies the identical one-line fix at the same spot (`axiosConfig.headers['Content-Type'] = undefined;`), plus proper regression tests and a release note. Closing this in favor of #2336, thanks for the quick turnaround!

PR #2364, opened 2026-09-15
feat(artifacts): one store for what a run produced, and keep chat images in it (#2362)

### LINKED ISSUE #2362: Images should be saved in persistence mode
When I generate an image in the persistence mode and I go back, the image is gone, but it should be kept.

manzke:
Closes #2362.

## The bug

A picture the model drew existed only in the tab that asked for it. `useChatMessages` strips image payloads before writing the transcript to `sessionStorage` — a generated image is megabytes and the quota is a few — so navigating away and coming back left the answer with an empty space where the image had been.

## The shape of the fix

An **artifact** is content a run produced that is worth keeping in its own right. A chat turn is one producer among several — a workflow's report and an agent's output are the same kind of thing — so this PR builds **one store** and puts the chat path on it, rather than bolting image storage onto chats.

`server/services/artifacts/ArtifactRepository.js` is addressed by a **scope**: `{ type: 'chat' | 'run', id }`, the thing whose lifetime the artifact follows. Keys are `<scopeType>__<scopeId>__<artifactId>`, so every artifact of one scope shares a prefix — which is what makes "everything this chat produced" and "everything this run produced" the same cheap walk, and what lets the sweep find an artifact whose descriptor never landed. `put` / `get` / `list` / `deleteMany` / `deleteScope` know nothing about chats.

`ChatRepository` is a caller: it takes the store as a dependency, exposes `artifactScope(chatId)`, and asks it to clean up when messages leave the transcript or the chat is deleted.

## Why the payload never goes in the transcript

A chat transcript and a workflow state document are single documents that their producer re-reads, re-serializes and re-hashes on every step, and that ship back whole when the thing is opened. One artifact inlined there is paid for on every step for as long as the producer lives. The producer records a descriptor:

```js
{ id: '7f3c…', kind: 'image', mimeType: 'image/png', bytes: 1483204 }
```

The client fetches the bytes per artifact through `apiClient`, not a plain `` — the URL is credentialed, and a bearer token in `localStorage` only travels on a request the client makes itself. Responses are `private, max-age=31536000, immutable`: an artifact document is written once, keyed by a fresh uuid, so what is behind a URL cannot change.

## API

| Method & path | Purpose |
| --- | --- |
| `GET /api/chats/:chatId/artifacts` | What this chat produced, newest first, descriptors only |
| `GET /api/chats/:chatId/artifacts/:artifactId` | The bytes of one, as its own media type |

Both authorize through the owning chat and answer **404** to anyone else, so an artifact id is never a capability on its own. The listing walks the artifact keys rather than the transcript, so it also sees an artifact whose descriptor never landed on a message.

## Kinds and content type

Media types are allowlisted **per kind**, so a new kind brings its own list and cannot widen another's:

- `image` — `png`, `jpeg`, `webp`, `gif`, `avif`, `bmp`, `heic`, `heif`
- `document` — `text/markdown`, `text/plain`, `text/csv`, `application/json`, `application/pdf`

Anything else is served `application/octet-stream` with `nosniff`. `image/svg+xml` and `text/html` are deliberately excluded: they are documents that can run script, and `nosniff` does not help when the type is honest. Adding a kind is an entry in `ARTIFACT_KINDS` plus a client renderer — the store, caps, endpoints, listing and cascade are all kind-agnostic.

## Configuration

A new `platform.json → artifacts` block, seeded by migration **V106** at its defaults (so an upgrade changes nothing):

| Key | Default | Meaning |
| --- | --- | --- |
| `enabled` | `true` | Master switch for every producer |
| `maxBytes` | `10485760` | Largest single artifact, base64 bytes; `<= 0` uncapped |
| `maxPerBatch` | `8` | Artifacts one producer records in one go; `<= 0` uncapped |

Deliberately **not** under `chats`: a workflow reading its own limits must not have to reach into the chat settings. An artifact a cap turns away is still described on the message (`{ kind, mimeType, bytes, unavailable }`) so a viewer can tell a dropped artifact from one that was never produced.

## What is NOT in this PR

The legacy agent artifact tree (`contents/data/agent-artifacts/<runId>/<name>`, written by `write_artifact` and the workflow executors, read by `GET /api/agents/runs/:runId/artifacts`) has **not** moved. It is not a drop-in swap — it stores text only, its read route hardcodes `text/markdown`, its quotas count against in-state bookkeeping rather than disk, and it has no retention at all while the workflow-state sweep deletes the state its authorization check depends on. Porting it is a data migration plus admin-UI, SSE-projection and tool changes, and belongs in its own change. `docs/artifacts.md` records exactly what it would take; `run` is already a scope and `document` is already a kind.

Workflow outputs are the other obvious producer: today a workflow's real deliverable (the Markdown/PDF/DOCX from the execution page) is generated in the browser and never persisted.

## Also in this PR

Merged `main`, which renamed the `thinkingBudget` chat setting to `thinkingLevel` (#2363), and renumbered the migration to V106 since main took 103–105.

## Testing

- `server/tests/artifact-repository.test.js` (new, 13 tests) — the shared store against a real filesystem provider: payload in its own document, two scopes kept apart, a scope that cannot be keyed is a no-op, the per-kind allowlist, listing order and payload-free entries, a scope id containing the key separator (listing *and* sweeping), emptying a scope including an unreferenced payload, the master switch, and storage-down degrading to no-ops.
- `server/tests/chat-persistence-artifacts.test.js` (9 tests) — the chat half: descriptors on the message and never a payload, the chat's scope, the delete/edit/message-cap cascades, the materializer's caps and refusal descriptors, and an answer that survives a failed artifact write.
- `server/tests/chat-persistence-routes.test.js` — six cases across both endpoints: listing as owner and as a stranger, bytes as owner, another owner, unknown artifact, traversal id.
- `tests/unit/client/generated-image.test.jsx` (7 tests) — hydration, live vs stored rendering, failure and refusal messaging, and the note only where true.
- `server/tests/migration-v106.test.js` — the block lands beside `chats`, not inside it; operator values including zero survive.
- Full `npm run test:quick` green (961 client unit tests + all server suites), `npm run lint` 0 errors, `npm run format` clean, and V106 verified applying against a live install.

## Documentation

- **`docs/artifacts.md`** (new, in `SUMMARY.md`) — the model, scopes, kinds, config, authorization, lifetime, how a new producer adopts it, and what has not moved yet.
- `docs/chat-persistence.md` — the chat's half, pointing at the store for the rest.
- `docs/storage.md` — the `artifacts` namespace.
- `docs/releases/5.5.0/features.md` — changelog entry.

https://claude.ai/code/session_01YaV8Awc71nBorAR1ZaGPPn


manzke:
### `Analyze (actions)` is red on `8869794`, and it is not this PR's

Everything else is green (`test-quick`, `test-integration`, `Server boots`, `Build Check`, `auto-lint-format`, `axe-core`, `label`, `Analyze (javascript-typescript)` ×2). Recording why I'm not chasing this one:

- **This PR touches no `.github/` files at all** — `git diff --name-only origin/main...HEAD` matches nothing under that path, so the Actions analysis has nothing new to look at.
- **It passed on the three earlier heads of this same PR** (`f111fc9`, `f526695`, `709123a`), against the same workflow files.
- **The analysis itself succeeded.** The job log shows all 18 `codeql/actions-queries` evaluated and the SARIF exported in 132 ms. What follows is `Uploading results`, a 20-second gap, then `CodeQL job status was failure` — with no error, no rule and no finding. The failing step is the code-scanning results upload, not the scan.
- GitHub's REST API was returning `500`/`502` on unrelated calls in that same minute, which fits an upload-side incident.
- The `CodeQL Advanced` workflow also has a long run of failures on `main`.

I would normally spend one re-run to confirm, but `rerun-failed-jobs` answers `403 This workflow run cannot be retried` for this session, so I can't. There is nothing to fix in the PR for it either — it isn't a test I could make more robust; the failure is in GitHub's upload path.

The PR stays watched. If this check is still red after the next push to it — one that *does* re-run the scan from scratch — I'll treat it as real and investigate again rather than repeat this.


---

PR #2423, opened 2026-09-17
feat(admin): configure and test the outbound proxy from the admin UI

### LINKED ISSUE #2417: Admin: configure and test the outbound HTTP(S) proxy from the admin UI
## Summary

The outbound HTTP(S) proxy — the corporate proxy iHub uses for its own egress to LLM providers, web search, Jira, OIDC, MCP servers — can only be configured by hand-editing `contents/config/platform.json` or by setting environment variables. There is no admin UI for it, no schema validation, and no way to verify a proxy actually works short of triggering real traffic and reading server logs.

This issue asks for two things:

1. **Configuration** of the outbound proxy as a first-class admin setting, alongside SSL/TLS and the SSRF allowlist.
2. **Testing** — a "Test connection" action that tells the admin, before and after saving, whether the configured proxy actually reaches a given destination, and *why* not when it doesn't.

### Scope clarification

This is about **egress proxying only**. Explicitly *not* in scope:

- `platform.json` → `proxyAuth` (header/JWT-based user authentication behind a reverse proxy, `auth.mode: "proxy"`). Already has its own admin surface.
- `platform.json` → `trustProxy` (how many inbound reverse-proxy hops Express trusts for `req.ip`).

Proxy *credentials* embedded in the egress proxy URL (`http://user:pass@proxy:8080`, documented in `docs/proxy-configuration.md`) **are** in scope — they are part of the outbound proxy config, not of `proxyAuth`.

## Current state

**Runtime support exists and works.** `server/utils/httpConfig.js` reads a `proxy` block and resolves it with this precedence (`getProxyConfig()`):

```
platform.json → config.env → process.env (HTTP_PROXY / HTTPS_PROXY / NO_PROXY)
```

It supports `enabled`, `http`, `https`, `noProxy`, and `urlPatterns` (regex allowlist for selective proxying), and `createAgent()` picks between a proxy agent and a direct agent per request. `TlsForwardingHttpsProxyAgent` works around an upstream `https-proxy-agent` bug so `rejectUnauthorized` reaches the destination TLS handshake.

**What's missing:**

| Gap | Detail |
| --- | --- |
| No admin UI | Nothing under `client/src/features/admin/` edits the egress proxy. `SSLConfig.jsx` and `SsrfConfig.jsx` are the obvious precedent — both are sections of `AdminSecurityPage.jsx`. |
| No schema | `server/validators/platformConfigSchema.js` has no top-level `proxy` key. The schema is `.passthrough()`, so the block survives round-trips, but it is never validated: a typo (`noproxy`, `httpsProxy`, a string where an array belongs) is accepted silently and the proxy just doesn't apply. It's also absent from `knownPlatformKeys`. |
| No admin API | There is no `/api/admin/proxy/config`, unlike `/api/admin/ssl/config` (`server/routes/admin/ssl.js`) and `/api/admin/ssrf/config`. |
| No test action | The repo already has the pattern — `POST /api/admin/models/:modelId/test`, `/api/admin/sources/:id/test`, `/api/admin/mcp/servers/:id/test`. Nothing equivalent for the proxy. |
| Credentials in plaintext | A proxy URL with embedded credentials sits unencrypted in `platform.json`, unlike Jira/OIDC/LDAP/NTLM secrets, which are AES-256-GCM encrypted at rest and redacted on admin read. |
| Config change is invisible | `getProxyConfig._logged` is a one-shot flag, so "Proxy configuration loaded" is emitted once per process. After an admin changes proxy settings there is no log confirming what is now in effect. (Related: #1088.) |

### Latent bugs: config that looks right but silently changes proxy behavior

Both of these are the same shape — a plausible config value is swallowed by a `try/catch`, the function falls through to a default, and the proxy quietly stops being used. Neither is visible without reading the source.

#### 1. One uncompilable `urlPatterns` entry stops the proxy being used

`matchesProxyPattern()` (`server/utils/httpConfig.js:286-301`) wraps the **entire** loop in a single `try/catch`:

```js
export function matchesProxyPattern(url, patterns) {
  if (!patterns || patterns.length === 0) return true; // no patterns → proxy everything

  try {
    for (const pattern of patterns) {
      const regex = new RegExp(pattern);   // throws on an uncompilable pattern
      if (regex.test(url)) {
        return true;
      }
    }
  } catch (error) {
    logger.warn('Error matching proxy pattern', { component: 'HttpConfig', error });
  }

  return false;   // ← reached after the catch
}
```

`new RegExp(pattern)` throws on a malformed pattern (`"api\\.openai\\.com("`, a stray `[`, a bad escape). The catch logs a warning and execution falls straight through to `return false`. Consequences:

- **Patterns from the bad entry onward are never evaluated, for any URL.** The loop aborts on the first throw; there is no per-pattern recovery.
- If the bad entry is **first in the list**, `matchesProxyPattern()` returns `false` for every URL, so the proxy is never used at all.
- If it is **later in the list**, every URL that doesn't match an earlier entry goes direct — including the URLs the remaining patterns were written to cover.
- `createAgent()` reads that `false` as "no proxy for this URL" and returns a direct agent (`server/utils/httpConfig.js:393-400`). In an environment where the proxy is the only egress path, the admin sees connect timeouts against `api.openai.com` and starts debugging the *destination*, not the typo in their own config.
- The warning doesn't name the offending pattern — it logs only `error`, so even reading the logs doesn't say which entry is broken. And a `warn` emitted per request is easy to miss entirely.

#### 2. A `noProxy` **array** silently disables every bypass

`shouldBypassProxy()` (`server/utils/httpConfig.js:241-277`) calls `noProxy.split(',')` *inside* its `try` block. `noProxy` is typed as a comma-separated string, but every other host list in `platform.json` is an array — `ssl.domainWhitelist` and `ssrf.allowedHosts` are both `z.array(z.string())`. An admin who reasonably writes

```json
"noProxy": ["localhost", "127.0.0.1", ".internal"]
```

gets a `TypeError: noProxy.split is not a function`, swallowed by the catch, logged as `Error parsing URL for proxy bypass` — and the function returns `false`. **Every** bypass stops working: internal hosts that should go direct are sent through the corporate proxy instead, which typically refuses them. The log line blames URL parsing rather than the config type, so it points the admin in the wrong direction.

See [§1 — `noProxy` type](#1-schema) for the proposed fix (accept both forms).

## Problem

Today, diagnosing a proxy problem means editing JSON on the server, restarting, generating real traffic, and grepping logs for `Using proxy for URL`. The failure modes an admin has to disambiguate by hand:

- proxy host unreachable (`ECONNREFUSED` / connect timeout) vs. destination unreachable through a working proxy
- `407 Proxy Authentication Required` — proxy needs credentials
- request silently bypassed the proxy because the host matched `noProxy`
- request silently bypassed the proxy because it did *not* match `urlPatterns`
- request silently bypassed the proxy because **another** `urlPatterns` entry failed to compile (latent bug 1) — indistinguishable from the previous case without reading source
- internal host went *through* the proxy because `noProxy` was written as an array (latent bug 2)
- TLS interception by the proxy (corporate MITM cert) rejected at the destination handshake
- `proxy.enabled` defaulting to `true` means an inherited `HTTPS_PROXY` in the environment applies without anyone having configured it in iHub

This is not hypothetical. #2415 was filed because an admin hit Brave Search timeouts, suspected the proxy, and had no repeatable way to test the proxy path in isolation. The answer shipped as manual `curl` instructions in `docs/proxy-testing-guide.md`. That should be a button in the admin UI.

### 1. Schema

Add a top-level `proxy` block to `server/validators/platformConfigSchema.js`, with `.describe()` on each field so it self-documents:

```js
proxy: z
  .object({
    enabled: z.boolean().prefault(true),
    http: z.string().prefault(''),
    https: z.string().prefault(''),
    noProxy: z
      .union([z.string(), z.array(z.string())])
      .prefault('')
      .describe(
        'Hosts that bypass the proxy. Accepts a comma-separated string (NO_PROXY format) or an array of entries.'
      ),
    urlPatterns: z.array(z.string()).prefault([])
  })
  .prefault({})
```

Validation should reject a `http`/`https` value that isn't a parseable `http(s)://host:port` URL, and reject `urlPatterns` entries that don't compile as regexes (`z.string().refine(p => { try { new RegExp(p); return true; } catch { return false; } })`), naming the offending entry in the error.

**`noProxy` type — accept both, normalize to an array.** The two forms exist for different reasons and both have to work:

- The **string** form can't go away. `noProxy` falls back to `process.env.NO_PROXY` (`httpConfig.js:213`), which is inherently comma-separated, and the shipped example config writes it as an env placeholder — `examples/config/platform.json:294` has `"noProxy": "${NO_PROXY}"`, where interpolation yields a string. Admins also copy-paste `NO_PROXY` values from other tooling.
- The **array** form is what the rest of `platform.json` looks like (`ssl.domainWhitelist`, `ssrf.allowedHosts`), what an admin UI list editor produces, and what avoids the `a, b,,c` whitespace handling the current `.split(',').map(trim)` papers over.

A union is additive: existing configs keep working, no migration, and no backward-compat question to put to the user. Normalize **once** in `getProxyConfig()` — split the string there and hand `shouldBypassProxy()` an array — so there is a single code path and the `@param {string} noProxy` JSDoc stops lying. That also fixes latent bug 2 by construction.

**Harden the runtime as well, not just the save path.** Schema validation only covers config written through the admin API; a hand-edited `platform.json`, a value from `config.env`, or an entry added before this lands still reaches these functions unvalidated. So also:

- move the `try/catch` in `matchesProxyPattern()` **inside** the loop so one bad pattern is skipped instead of aborting evaluation of the rest
- include the offending pattern string in the warning so the log identifies the culprit
- consider compiling `urlPatterns` once when the config is loaded rather than per request, which turns this into a single startup/save-time warning instead of a per-request one

Whether a fully-unparseable pattern list should fail open (direct) or fail closed (proxy everything) is worth a decision in review — today it fails open silently, which is the behavior that misleads.

**Three `NO_PROXY` semantics gaps, while `shouldBypassProxy()` is open.** Each is small; the point is to decide explicitly rather than inherit them. If they aren't implemented, `docs/proxy-configuration.md` should say so, because `curl` and Node's own handling accept all three:

- CIDR notation and IP ranges — there's a comment at `httpConfig.js:271` saying they're deliberately unsupported
- a bare `*` meaning "bypass everything"
- `host:port` entries (e.g. `internal.example.com:8443`)

No migration is needed: the block is optional, absent-means-unset is already the runtime behavior, and `server/defaults/config/platform.json` ships no egress `proxy` block. (If a default `proxy: {}` is added there, that *is* a migration — use the `create-migration` skill.)

### 2. Admin API

`server/routes/admin/proxy.js`, modeled on `ssl.js`:

- `GET /api/admin/proxy/config` — returns the effective config **plus its provenance** per field (`platform.json` / `config.env` / `process.env` / unset), with `noProxy` normalized to an array so the client has one shape to render. Provenance matters: an admin looking at an empty form needs to know an env var is proxying everything anyway. Credentials in proxy URLs redacted on read (`redactUrlSecrets()` already exists in `httpConfig.js`).
- `PUT /api/admin/proxy/config` — validate, write, `configCache.refreshCacheEntry('config/platform.json')`. Since `getProxyConfig()` reads the cache per call, saved changes take effect without a restart; confirm this end-to-end and state it in the UI.
- Log the effective config after a successful save (fixes the `_logged` one-shot gap above).

Decide and document whether `PUT` writes `noProxy` back as an array or preserves the form it found — round-tripping a `${NO_PROXY}` placeholder into a literal array would silently break env-driven deployments, so a placeholder value must survive a save untouched.

### 3. Admin UI

A `ProxyConfig` section on `AdminSecurityPage.jsx` (next to SSL / CORS / SSRF), or a new "Network" settings page if Security is getting crowded. Fields per the schema, with:

- inline provenance hints ("currently set by `HTTPS_PROXY` environment variable")
- a note that `enabled: true` is the default, so env-var proxies apply unless explicitly disabled
- `noProxy` and `urlPatterns` as list editors, mirroring the `domainWhitelist` editor in `SSLConfig.jsx`; pasting a comma-separated string into the `noProxy` editor should split into entries rather than being stored as one
- live regex validation on `urlPatterns` — an entry that doesn't compile is flagged in the editor, before it can take the proxy down
- a hint that `urlPatterns` is an allowlist: empty means "proxy everything", non-empty means "proxy only what matches"
- all strings via i18n (`shared/i18n/en.json` + `de.json`)

### 4. Test endpoint — the core of this request

`POST /api/admin/proxy/test`, accepting a target URL and **optionally a draft (unsaved) proxy config**, so an admin can validate settings *before* committing them.

The response should answer "what happened", not just pass/fail:

- **Routing decision** — which branch of `createAgent()` the URL took: proxied / bypassed via `noProxy` / excluded by `urlPatterns` / excluded because a `urlPatterns` entry failed to compile / direct because no proxy configured. This alone resolves the silent-bypass failure modes.
- **Proxy used** — the proxy URL with credentials redacted.
- **Outcome** — HTTP status, total duration, and split timings (connect-to-proxy vs. end-to-end) where feasible.
- **Classified error on failure**, distinguishing at minimum: proxy unreachable (`ECONNREFUSED`/`ETIMEDOUT` against the proxy host), `407` proxy auth required, DNS failure at the destination, TLS failure at the destination (surfacing the cert chain issuer so corporate MITM interception is recognizable), destination HTTP error.
- **Suggested next step** per classification — e.g. for a TLS failure, point at `ssl.domainWhitelist`; for a bypass, point at the `noProxy`/`urlPatterns` entry that matched, or at the pattern that failed to compile.

Targets: a small curated list of the endpoints that actually matter (configured LLM provider base URLs, the web-search endpoint, configured MCP servers) plus a free-form URL field. The endpoint is `adminAuth`-gated, but since the target URL is caller-supplied it is an SSRF surface — it must go through the existing guard and honor `ssrf.allowedHosts` rather than fetching arbitrary internal addresses, and the test must not follow redirects or return the response body (status, headers, and timing only).

Surface it in the UI as a "Test connection" button with a results panel, reusing `IntegrationTestResults.jsx` if it fits.

### 5. Secrets at rest

Route the `http`/`https` proxy URLs through `TokenStorageService` like other platform secrets: encrypt on admin save, decrypt + redact on admin read, decrypt into cache at runtime. Plaintext already in `platform.json` passes through and is lazily encrypted on next save — the established pattern. Redaction must also hold in the test endpoint's response and in any log line that echoes a proxy URL.

## Acceptance criteria

- [ ] `proxy` is a validated block in `platformConfigSchema.js`; invalid proxy URLs and uncompilable `urlPatterns` are rejected with a clear message naming the offending entry
- [ ] `noProxy` accepts both a comma-separated string and an array, is normalized once in `getProxyConfig()`, and an array value produces working bypasses (regression test for latent bug 2)
- [ ] A `${NO_PROXY}`-style placeholder in `noProxy` survives an admin save unchanged
- [ ] A single uncompilable `urlPatterns` entry no longer aborts evaluation of the remaining patterns, and the warning names the pattern that failed
- [ ] `GET`/`PUT /api/admin/proxy/config` exist, are `adminAuth`-gated, Swagger-documented, and report per-field provenance
- [ ] Saved proxy changes take effect without a server restart (verified end-to-end, not just assumed from the cache read path)
- [ ] An admin can view and edit `enabled` / `http` / `https` / `noProxy` / `urlPatterns` in the admin UI, fully i18n'd (en + de); `noProxy` and `urlPatterns` are list editors and the pattern editor flags entries that don't compile
- [ ] Proxy credentials are encrypted at rest and never returned unredacted by an admin endpoint or written to a log
- [ ] `POST /api/admin/proxy/test` reports the routing decision, the proxy used, status and timing, and a classified error with a suggested next step — and runs against draft config as well as saved config
- [ ] The test endpoint honors the SSRF guard, does not follow redirects, and does not return response bodies
- [ ] The effective proxy config is logged after every successful save
- [ ] Unit tests for: the new schema validation, `noProxy` normalization from both forms, `matchesProxyPattern()` with a bad entry first / in the middle / last (later patterns still evaluated in each case), the routing-decision classifier, and the error classifier; the existing `server/tests/httpConfig.proxy-tls.test.js` still passes
- [ ] `docs/proxy-configuration.md` documents the admin UI path, both accepted `noProxy` forms, the allowlist semantics of `urlPatterns`, and which `NO_PROXY` syntaxes are *not* supported; `docs/proxy-testing-guide.md` leads with the built-in test and keeps the manual `curl` steps as a fallback
- [ ] Changelog entry in `docs/releases/next/` (`/document-feature`)

## Out of scope

- `proxyAuth` / `auth.mode: "proxy"` (inbound header-based user authentication)
- `trustProxy` (inbound reverse-proxy hop count)
- PAC file support and SOCKS proxies — worth a separate issue if there's demand
- Per-app or per-model proxy overrides; this stays a platform-level setting

## References

- `server/utils/httpConfig.js` — `getProxyConfig()`, `createAgent()`, `shouldBypassProxy()`, `matchesProxyPattern()`, `redactUrlSecrets()`
- `server/routes/admin/ssl.js`, `server/routes/admin/ssrf.js` — the API pattern to mirror
- `client/src/features/admin/components/SSLConfig.jsx`, `SsrfConfig.jsx`, `AdminSecurityPage.jsx` — the UI pattern to mirror
- `server/routes/admin/mcpServers.js` — existing "test connection" pattern
- `examples/config/platform.json:290-296` — the shipped `proxy` block
- `docs/proxy-configuration.md`, `docs/proxy-testing-guide.md`
- `concepts/2025-11-20 Proxy Support for External Services.md`
- #560 (proxy support, implemented), #1088 (print proxy configuration, closed), #2415 (manual Brave Search proxy testing steps — the trigger for this issue)

manzke:
Closes #2417

The outbound proxy iHub uses for egress — LLM providers, web search, Jira, OIDC, MCP servers — could only be changed by hand-editing `contents/config/platform.json` or the environment, with no schema validation and no way to tell whether the change had landed. It is now a first-class setting under **Admin → Security → Outbound Proxy**, next to SSL/TLS and the SSRF allowlist, with a connectivity test.

## Latent bugs fixed

Two ways of writing a valid-looking `proxy` block silently disabled proxying altogether, with nothing in the log to say so:

- **One uncompilable `urlPatterns` entry stopped all pattern evaluation.** `matchesProxyPattern()` wrapped the whole loop in a single `try/catch`, so a bad entry aborted evaluation of every entry after it — and a bad *first* entry meant no URL ever matched, sending everything direct. Each pattern is now compiled on its own; a bad one is skipped with a warning naming it (once, not per request) and the rest still apply.
- **A `noProxy` array broke every bypass.** `shouldBypassProxy()` called `String.split()` on it, so an array — the shape the neighbouring `ssl.domainWhitelist` uses — threw a `TypeError` into a swallowed `catch` and all bypasses stopped working. Both the comma-separated string and the array are now accepted and normalized to an array once, in `getProxyConfig()`.

Also: an `${ENV_VAR}` placeholder that no variable resolves is no longer handed to the proxy agent as if it were an address (the literal `"${HTTPS_PROXY}"` is truthy). The connection goes direct instead of failing on an unparseable URL.

## Server

- **Schema** (`platformConfigSchema.js`): new exported `proxyConfigSchema` validating `enabled` / `http` / `https` / `noProxy` / `urlPatterns`. Proxy URLs must parse as absolute http(s) URLs, be an `${ENV_VAR}` placeholder, or be empty; `noProxy` is a union of string and array; each `urlPatterns` entry must compile as a regex. Failures name the offending field and pattern.
- **`GET`/`PUT /api/admin/proxy/config`**: `adminAuth`-gated and Swagger-documented. Reports what is stored, what is actually in effect, and per-field provenance (`platform` / `environment` / `default`). Passwords are masked as `***REDACTED***`; sending the mask back keeps the stored password while still allowing host/port/user edits. URLs are encrypted at rest via `TokenStorageService`; `${ENV_VAR}` placeholders are stored verbatim and survive saves. The cache is refreshed so changes apply without a restart, and the effective config is logged afterwards.
- **`POST /api/admin/proxy/test`**: probes one URL against the saved config or an unsaved draft. Returns the routing decision (`proxied` / `bypassed` / `excluded` / `disabled` / `direct`), whether the proxy answers at TCP level, the HTTP status, timings split between proxy connect and request, and a classified failure with a suggested next step (`proxy_unreachable`, `proxy_auth_required`, `proxy_dns_failure`, `dns_failure`, `tls_failure`, `timeout`, `target_unreachable`, `http_error`, `request_failed`). It honors `ssrf.allowedHosts`, applies the same SSL decision the real request would, follows no redirects and returns no response body.
- **`describeProxyRouting()`** is now the single source of truth for the routing rules, shared by `createAgent()`, the integration diagnostics and the test endpoint, so the three cannot disagree. Proxy URLs are redacted in the startup log.

## Client

New `ProxyConfig` section on the Security page: enable switch, HTTP/HTTPS URL fields with provenance hints and a warning for unresolved placeholders, list editors for `noProxy` and `urlPatterns` with live regex validation, an "in effect right now" summary, and the test panel. Fully translated (en + de), including the `admin.security` section labels the page had been falling back to English for.

## Migration, tests, docs

- `V110__add_proxy_defaults.js` seeds the block. `enabled` is seeded `true`, not `false`: `getProxyConfig()` has always treated an absent flag as enabled, so writing `false` would silently cut egress for installs configured purely through `HTTP_PROXY`/`HTTPS_PROXY`.
- `server/tests/httpConfig.proxy-routing.test.js` (27 tests) — `noProxy` normalization in both forms, bypass semantics, a bad `urlPatterns` entry in first/middle/last position, placeholder and `ENC[...]` handling, provenance, and every routing decision.
- `server/tests/admin-proxy-config.test.js` (25 tests) — masking and encryption round-trip, placeholder preservation (including the case where the cache holds the resolved value), validation rejections, the SSRF refusal, draft evaluation, and the failure classifiers.
- `docs/proxy-configuration.md` and `docs/server-config.md`: the admin UI path, both `noProxy` forms with a semantics table, the unsupported `NO_PROXY` syntaxes (CIDR, `host:port`, bare `*`), `urlPatterns` semantics, encryption at rest, and the API/classification reference. Changelog entries in `docs/releases/next/`.

## Verification

- `server` suite: same 50 pre-existing failures as on `main`, 790 passing (up from 724) — no regressions.
- `npm run test:unit`: 1115 passing, 100 suites.
- Lint and Prettier clean on every changed file; client builds; server boots and applies V110.
- Exercised end-to-end against a running server: encrypted-at-rest save, masked round-trip with a changed host, validation rejections, and the test endpoint against a real proxy (200 through the proxy), an unreachable proxy, a `noProxy` bypass, a `urlPatterns` exclusion and an SSRF-blocked target. Checked the admin UI in a browser: the `${HTTPS_PROXY}` placeholder survives a load, invalid regexes are flagged live, and the test panel renders the routing decision, proxy reachability, split timings and next step.

https://claude.ai/code/session_01FTxCZaYftPFPwy8H1kaDCp

---
_Generated by [Claude Code](https://claude.ai/code/session_01FTxCZaYftPFPwy8H1kaDCp)_

---

## Known CodeQL alert (accepted by design)

The **CodeQL** check is red on one alert — `js/request-forgery` (critical) on the outbound call in `POST /api/admin/proxy/test`. **This is intended behaviour, not a defect, and the maintainer decision is to dismiss it** in the Security tab as "used in tests / intended by design". Everything else on the head is green.

Probing an admin-supplied URL is what the endpoint is for (#2417: "Accepts target URL and optional draft config"). The call is constrained by `adminAuth`, http(s) schemes only, `assertSafeHost()` rejecting anything that resolves into a private range unless it is listed in `ssrf.allowedHosts`, no redirect following, and a response body that is never read or returned.

It cannot be resolved in code without narrowing that contract: CodeQL stops flagging only when the request URL's origin comes from server-side configuration instead of the request body — which is how `admin/integrationTest.js` avoids it, since it probes URLs read from `configCache`. `// codeql[...]` markers are not honoured by GitHub code scanning, and CodeQL here runs via GitHub's default setup, so it cannot be configured from the repository. The two other alerts from the first run were real and are fixed (see `37f77dd` and `6d7b4ae`).

---


manzke:
### CodeQL `js/request-forgery` (critical) is blocking this PR

**What is failing:** the CodeQL check on `6d7b4ae`. One alert remains, on the outbound call in `POST /api/admin/proxy/test`: *"The URL of this request depends on a user-provided value."*

The other two alerts from the first run are fixed and gone:

- **Disabling certificate validation** — the endpoint no longer builds its own agent. `createAgent()` now takes a proxy-config override, so the test goes through the same call every outbound request makes.
- **Resource exhaustion** — the caller-supplied `timeoutMs` is clamped in `clampTestTimeout()` with explicit comparisons instead of `Math.min`/`Math.max`.

**Why the last one is not a code defect:** probing an admin-supplied URL is what this endpoint is *for* — issue #2417 specifies "Accepts target URL and optional draft config". The call is already constrained: `adminAuth`, http(s) schemes only, `assertSafeHost()` rejecting anything that resolves into a private range unless it is in `ssrf.allowedHosts`, no redirect following, and the response body is never read or returned.

**What I tried that did not work:** a `// codeql[js/request-forgery]` marker matching the one in `server/routes/admin/tools.js`. GitHub code scanning does not honor those markers — the two alerts that cleared did so because the offending code was removed, not because of a marker. Routing the probe through `httpFetch()` (correct on its own merits, since `httpConfig.js` states every outbound call should use it) did not clear it either: CodeQL follows the taint across modules.

**Why there is no in-scope code fix:** CodeQL only stops flagging this when the request URL's origin is untainted — i.e. when the target comes from server-side configuration rather than the request body. That is exactly how `admin/integrationTest.js` avoids the alert today: it probes URLs read from `configCache`, never from `req.body`. Any CodeQL-clean version of this endpoint therefore has to stop accepting an arbitrary URL, which contradicts the issue's acceptance criteria. CodeQL here runs via GitHub's *default setup*, so its configuration cannot be changed from the repository either.

**Two ways forward — this needs a maintainer decision:**

1. **Dismiss the alert** as intended-by-design in the Security tab. Keeps the endpoint as specified in #2417. Reasonable given it is admin-only, private-IP-guarded, returns no body and follows no redirects.
2. **Restrict the target** to origins iHub is already configured to reach (models, sources, MCP servers, Jira, OIDC, web search) plus `ssrf.allowedHosts` entries, with the submitted URL used only to select a config-derived target. CodeQL-clean by construction and a smaller attack surface, but "test any URL" is gone — an operator wanting to try `https://example.com` would first have to add the host under Security → SSRF Allowlist.

I have not implemented option 2, because narrowing a documented contract is the author's call, not mine. Say which you want and I will push it — option 2 is roughly a day's-worth-smaller change than it sounds: a config-derived target list, an origin match, and a dropdown in the admin panel.

The rest of the PR is unaffected: every other check is green and the branch is mergeable.

---

manzke:
**Decision: option 1 — dismiss the alert as intended by design.** The endpoint stays as specified in #2417; no further code change.

To clear the check, a maintainer needs to dismiss alert `js/request-forgery` on `server/routes/admin/proxy.js` under **Security → Code scanning**, with reason *"Used in tests" / "Won't fix — intended by design"*. I do not have code-scanning write access (the API returns 403 for this token), so I cannot action it myself.

The rationale is now recorded in the PR description so it is not lost when this thread scrolls. Every other check on the head is green and the branch is mergeable.

---

PR #2425, opened 2026-09-18
Feat/feedback visibility toggle

### LINKED ISSUE #2424: Admin: one place to configure and review feedback (global + per-app toggle)
## Summary

Make feedback collection something an admin controls and reviews in one place:

- a **global** on/off switch for the response-feedback UI and API,
- a **per-app** override so individual apps can opt out,
- a single **Admin → Feedback** page that holds the feedback configuration (collection, storage, identification mode, per-app overrides) *and* the submitted feedback for review.

## Problem

Three separate problems, all about the same feature:

1. **No way to turn it off.** Feedback controls are hard-coded into `ChatMessage.jsx` and always visible in every chat surface (main chat, compare mode, canvas, Office add-in). The only related flag, `features.feedbackTracking`, controls server-side **storage** of already-submitted feedback — it does not hide the UI or block submission.
2. **No per-app control.** An app whose answers are machine-generated, internal, or regulated may not want a star rating under every response, even where feedback is fine platform-wide.
3. **Configuration and review live apart.** `feedbackTracking` and `usageTrackingMode` (pseudonymisation) are only editable by hand-editing `contents/config/platform.json` — or, for the tracking mode, buried in Usage Reports — while the submitted feedback is reviewed in Admin → Usage Reports → Feedback. Nothing in the product presents "feedback" as one thing.

### 1. Global toggle (Feature Registry)

Registry entry in `server/featureRegistry.js`: `id: 'feedback'`, `category: 'content'`, `default: true`. The generic Features page keeps rendering it automatically; the new Feedback page surfaces the same flag through the existing `/api/admin/features` endpoint, so there is one stored value, not two.

### 2. Per-app override (app config)

`features.feedback: boolean` on the app config (`server/validators/appConfigSchema.js`), defaulting to enabled when absent — the same shape as the existing `features.compareMode`/`features.magicPrompt` app flags.

- **Client**: `useFeatureFlags().isBothEnabled(app, 'feedback', true)` in `ChatMessage.jsx` — platform AND app must allow it. `ChatMessage` already receives `app` from every surface that renders it (`ChatMessageList`, `ComparePanel`, `CanvasChatPanel`, `OfficeChatPanel`, `AppChat`), so a single check there covers all of them; no prop threading needed.
- **Server**: `POST /api/feedback` gets `requireFeature('feedback')` in its middleware chain plus a per-app check on the request's `appId`, both returning the shared `403 { error, code: 'FEATURE_DISABLED' }`. The API cannot be used to bypass either level.
- **Admin app editor**: a toggle in the app form's features section, so the override is set where the rest of the app's features are.

### 3. Admin → Feedback: one place (`/admin/feedback`)

A new admin page under **Observability**, with two tabs:

**Settings**
- *Collect feedback* — the global `feedback` flag (`/api/admin/features`).
- *Per-app overrides* — a table of apps with a toggle each, so an admin can switch feedback off for single apps without opening every app editor. Reads `/api/admin/apps`, writes through the existing app update endpoint.
- *Store feedback* — `features.feedbackTracking` (visibility vs. persistence stay independent; the page explains the difference instead of hiding it).
- *Identification mode* — `features.usageTrackingMode` (anonymous / pseudonymous / identified) through the existing `/api/admin/usage/meta` endpoint, clearly labelled as shared with usage tracking.

**Feedback (review)**
- The rating overview (distribution, average, totals) and the entries browser with comments and pagination, moved over from Usage Reports → Feedback.

Usage Reports keeps its *Average Rating* tile, now linking to the new page; its **Feedback tab is removed** so there is exactly one place for feedback. No data is moved, only the UI. No new top-level route, so `KNOWN_ROUTES`/`index.html` need no change.

### 4. No config migration

The registry resolves an absent key to its declared `default` at read time, and the per-app flag is optional — existing installations behave exactly as before.

## Breaking Changes

**None for end users or the API.** Both switches default to enabled, so nothing changes until an admin opts out.

Admin-visible relocation: the feedback review moves from Admin → Usage Reports → Feedback to Admin → Feedback. No action required during upgrade; the Usage Reports average-rating tile links to the new location.

## Acceptance Criteria

- [ ] Admin → Feedback exists, holding both configuration and the submitted feedback
- [ ] Global toggle off: star-rating control + feedback modal hidden under every AI response, in every chat surface (main chat, compare mode, canvas, Office add-in)
- [ ] Global toggle off: `POST /api/feedback` → `403 FEATURE_DISABLED`
- [ ] Per-app override off (global on): feedback hidden in that app only; `POST /api/feedback` for that `appId` → `403 FEATURE_DISABLED`; other apps unaffected
- [ ] Per-app override settable both from Admin → Feedback and from the app editor, and absent from an app config means enabled
- [ ] The same global toggle is visible and consistent in Admin → Features → Content and Admin → Feedback (one stored value)
- [ ] Storage (`feedbackTracking`) and identification mode (`usageTrackingMode`) are editable from Admin → Feedback, with their scope explained
- [ ] Re-enabling immediately restores UI and API — no restart required
- [ ] Already-submitted feedback stays readable in Admin → Feedback regardless of either flag's current state
- [ ] `feedbackTracking` behaviour is unchanged by the new flags (visibility vs. persistence stay independent)
- [ ] `docs/feedback-feature.md` documents both switches and the new admin location

## Out of Scope

- Per-group overrides (global + per-app only for this iteration)
- Changing how feedback is stored, retained, or deleted (no retention/erasure UI)
- Changing `feedbackTracking`'s or `usageTrackingMode`'s own semantics
- An end-user explanation of why feedback is unavailable

## Areas Touched

`server/featureRegistry.js`, `server/routes/chat/feedbackRoutes.js`, `server/validators/appConfigSchema.js`, `client/src/features/chat/components/ChatMessage.jsx`, `client/src/features/admin/pages/AdminFeedbackPage.jsx` (new), `client/src/features/admin/pages/AdminUsageReports.jsx`, `client/src/features/admin/components/AdminSidebarNavData.js`, `client/src/features/admin/components/AppFormEditor.jsx`, `client/src/App.jsx`, `docs/feedback-feature.md`

---

jwalteri:
Closes  #2424

## Summary

Admins can now disable response feedback (star rating + comment) platform-wide from **Admin Panel → Features → Content**. Until now the star-rating control under AI responses was always on with no supported way to hide it — the only existing related flag, `features.feedbackTracking`, only stops server-side storage, not UI visibility or submission.

## Solution

Reuses the existing Feature Registry end-to-end, no new subsystem:

- **`server/featureRegistry.js`**: new `feedback` entry (`category: 'content'`, `default: true`). The generic admin Features page already renders and persists any registry entry, so no new admin UI was needed.
- **`server/routes/chat/feedbackRoutes.js`**: `POST /api/feedback` gated with `requireFeature('feedback')`, the same pattern already used for `tools`/`sources`/`skills`. Returns the shared `403 { error, code: 'FEATURE_DISABLED' }`.
- **`client/src/features/chat/components/ChatMessageList.jsx`** resolves the flag inline (`useFeatureFlags().isEnabled('feedback', true)`, matching how every other platform-level flag is checked elsewhere) and threads it as one boolean prop into **`ChatMessage.jsx`**, gating both the star-rating row and the feedback modal off that single prop. Because `ChatMessage` is the one shared component every chat surface renders through, this covers main chat, compare mode, canvas, and the Office add-in without surface-specific changes.
- No config migration: the registry resolves a missing key to its `default` at read time.

## Breaking Changes

None. Default is `true`, so no existing installation's behavior changes until an admin opts out. The pre-existing `features.feedbackTracking` flag is untouched and stays independent.

## Fixes found while verifying this end-to-end

A `/code-review high` pass plus manual Docker verification surfaced three issues, all fixed in this PR:

- **`docs/feedback-feature.md`** pointed admins at the wrong config file for the new flag (`platform.json` instead of the actual `features.json`) and its Troubleshooting section didn't cover the new `403` failure mode — both corrected.
- A dedicated `useFeedbackEnabled()` hook claimed to follow an "established one-hook-per-flag idiom" that doesn't actually exist anywhere else in the codebase — removed in favor of the inline `isEnabled(...)` call every other flag consumer uses.
- `ChatMessage.jsx`'s `showFeedbackForm` state wasn't reset when the flag disabled the modal, so it could silently reappear if the flag flipped back on before the user closed it another way — fixed with a `useEffect`.
- Unrelated, pre-existing bug surfaced during manual testing: `POST /api/feedback` rejected every normal (non-iFinder) submission with `400`, because `feedbackSchema` declared `conversationId`/`ifinderMessageId` as `z.string().optional()`, which Zod v4 doesn't accept `null` for — but the client always sends `null` when there's nothing to report. Fixed with `.nullable().optional()`, matching the existing pattern in `auditEntrySchema.js`.

Full rationale for both the design and every finding (including what was deliberately *not* changed) is in [`concepts/2026-09-17 Feedback Visibility Toggle.md`](concepts/2026-09-17%20Feedback%20Visibility%20Toggle.md).

## Testing

- `server/tests/feedback-visibility.test.js`: `POST /api/feedback` → 403 disabled / 2xx enabled / 2xx on a missing key (default)
- `tests/unit/client/chat-message-feedback-visibility.test.jsx`: star rating + modal present/absent per flag, never for user messages, default-enabled when the prop isn't threaded
- `tests/unit/server/feedbackSchema.test.js`: regression test for the null-handling fix
- `npm run test:quick` — full suite green
- `npm run lint:fix && npm run format:fix` — clean
- Manually verified in Docker: toggle in Admin Panel hides/shows the control live, disabled flag returns 403 via curl, end-to-end feedback submission works after the schema fix

🤖 Generated with [Claude Code]


manzke:
will redo it

PR #2427, opened 2026-09-18
feat(oauth): govern MCP clients identified by a metadata document

### LINKED ISSUE #2421: [MCP] Client governance — block clients, per-client group/app policy and revoke all connections
## Background

Since #2358 claude.ai connects to the MCP gateway as a **CIMD** client: its `client_id` is the HTTPS URL of a metadata document it publishes, and iHub stores no record for it. Admin → OAuth → Clients shows such clients as **synthetic, read-only rows** derived from the consent store (`listSeenCimdClients()`), and Admin → OAuth → Connections can revoke **one user's** grant for **one client**.

Governance stops there. What an administrator can control, per client kind:

| Control | Stored client (admin / DCR) | CIMD client (what Claude uses) |
|---|---|---|
| Suspend / block | ✅ `active: false` | ❌ only by disabling CIMD or dropping the whole host |
| Allowed groups | ✅ `allowedGroups` | ⚠️ one global list for **all** CIMD clients |
| Allowed apps / models / prompts | ✅ per client | ⚠️ one global list for **all** CIMD clients |
| Grantable scopes / token lifetime | ✅ per client | ⚠️ one global value |
| Revoke one user's connection | ✅ | ✅ |
| Revoke **all** connections of a client | ❌ | ❌ |
| Approve a client before it may connect | ❌ | ❌ (host allowlist only) |

Four concrete problems.

**P1 — Host granularity is the wrong granularity.** `oauth.cimd.allowedClientHosts` works on hostnames, but Claude web, Claude Desktop, Claude Code and Cowork all publish their documents under `claude.ai` with *different* `client_id` URLs. "Only the `claude-code` group may connect Claude Code, and nobody may connect Claude web" is not expressible — allowing the host allows all of them, and a host that is allowed today silently covers every client that vendor publishes tomorrow.

**P2 — Everything CIMD shares one policy.** `getCimdConfig()` (`server/utils/oauthClientResolver.js:29`) reads a single `platform.oauth.cimd` block and `buildCimdClient()` stamps `allowedGroups`, `allowedApps`, `allowedModels`, `allowedPrompts`, `scopes` and `tokenExpirationMinutes` from it onto *every* CIMD client. Narrowing Claude Code to two apps narrows every other metadata-document client to the same two apps.

**P3 — No bulk revocation and no way to block.** `revokeConnection(clientId, userId)` is per pair. Retiring or cutting off a client means clicking every row on the Connections page — and the user is reconnected the moment they re-authorize, because nothing marks the client as unwelcome. There is no `active: false` to set: a CIMD client has no record.

**P4 — Policy is enforced at consent time, not on the request path.** `oauthAuthorize.js` checks `allowedGroups` on GET and on the consent POST (`isUserAllowedByGroups`, line 119). But the **refresh grant** in `oauth.js` re-checks `refreshClient.active` only, and rebuilds the user from `tokenData.userGroups` — the group snapshot frozen into the refresh-token entry at first authorization (`server/routes/oauth.js:456-462`), which rotation re-stamps unchanged. `mcpAuth.js` never checks `allowedGroups` at all.

> **Net effect: removing a user from the allowed group does not end their MCP access.** Their client keeps rotating refresh tokens with the original group list until an administrator revokes the connection by hand. Same for narrowing a client's `allowedGroups`.

Full design with rationale, alternatives and the honest limits: ``['concepts/2026-09-17 MCP Client Governance — Per-Client Policy, Blocking and Connection Revocation.md'](https://github.com/intrafind/ihub-apps/blob/claude/modest-maxwell-n81s7d/concepts/2026-09-17%20MCP%20Client%20Governance%20%E2%80%94%20Per-Client%20Policy%2C%20Blocking%20and%20Connection%20Revocation.md).`` Read it before implementing.

## Goal

- **Revoke** every connection of a client in one action, and have a block keep it revoked — the client must re-authenticate and re-consent, or be refused outright.
- **An unknown client does not connect.** Passing the host allowlist makes a client *eligible*, not *allowed*; until an administrator approves it, the authorization is refused with a page telling the user to ask for it.
- **Block** specific clients by `client_id` URL, and by host pattern for a blunter cut.
- **Per-client resource policy**: `allowedApps`, `allowedModels`, `allowedPrompts`, grantable `scopes`, `tokenExpirationMinutes` — the fields stored clients already have.
- **Per-client group policy**: "only `claude-code-users` may connect Claude Code."
- Policy changes take effect on the **request path**, bounded by the access-token lifetime.

## Non-goals (follow-ups)

- Re-reading group membership from an external IdP mid-session (see *Honest limits*).
- RFC 8707 resource indicators / audience binding — still open from #2358.
- Per-tool policy on the gateway (tool visibility stays derived from accessible apps, `server/services/mcp/permissions.js`).
- Rate limiting or quota per client; shortening the `refreshTokenExpirationDays` default.

## Design principles

1. **Identity from the document, policy from the record.** Policy-only records live in the existing `contents/config/oauth-clients.json`, keyed by the CIMD `client_id` URL, with `metadata.cimd: true` and **no secret**. `client_name`, `redirect_uris`, `grant_types` and `token_endpoint_auth_method` keep coming from the document, never from the record — CIMD's "nothing stored" property is about identity and stays intact. Only administrator decisions are stored, as they always have been.
2. **Reuse the enforcement that exists.** `mcpAuth` already refuses `!client.active`; `oauthAuthorize` already refuses on `allowedGroups`; `authorization.js` already intersects `clientAllowedApps` into the user's permissions (`intersectWithClientAllowList`, line 496). Blocking and per-client policy become *data* in paths that already enforce them.
3. **Immediate, cluster-wide.** `loadOAuthClients()` reads through `configCache` and `saveOAuthClients()` announces an invalidation, so a block lands on every worker with no restart. (A `platform.json` home for this would need one — wrong for a kill switch.)
4. **Field-by-field layering.** `effective(field) = record[field] ?? platform.oauth.cimd[field]`. An admin who set a global `allowedApps` and then narrows `allowedGroups` on one client keeps the global apps list on that client.
5. **One resolver, one client shape.** Both `resolveOAuthClient()` (authorize, may fetch) and `buildPolicyCimdClient()` (request/token paths, never fetches) apply the same layering, so the gateway and the authorize endpoint cannot disagree. Downstream code still must not branch on `kind`.
6. **CIMD clients are still never trusted.** `consentRequired: true` and `trusted: false` are locked on these records, writable by no path. Approval is not trust — an approved client still sends every user through sign-in and consent.
7. **`active` is computed on every request**, from five independently revocable conditions:
   `cimd.enabled && hostAllowed && !hostBlocked && record.active !== false && approvalSatisfied`.

## Implementation plan

Three PRs. Phase 1 delivers the operational ask on its own; Phase 3 turns the gate on.

### Phase 1 — Revocation and blocking

- [ ] `server/services/oauth/ConnectionService.js`: `revokeConnectionsForClient(clientId)` — every consent and every refresh token for the client, built on the existing `revokeConnection` so there is one definition of revoking; returns `{ connectionsRevoked, refreshTokensRevoked }`.
- [ ] `server/routes/admin/oauthClients.js`: `DELETE /api/admin/oauth/clients/:encodedClientId/connections` (adminAuth, `logAudit`). Client ids are **base64url-encoded** in the path — `validateIdForPath()` rejects URL ids, the same obstacle `oauthConnections.js` already works around by treating the value as a store key, not a path segment.
- [ ] `server/utils/oauthClientManager.js`: `upsertCimdClientPolicy(clientId, patch, savedBy)` — writes a secretless record under the exact URL key; refuses `clientSecret`, `trusted`, `consentRequired`. (`createOAuthClient()` cannot be reused: it mints a UUID id and a bcrypt secret.)
- [ ] `server/routes/oauthAuthorize.js`: on a successful CIMD authorization, write a discovery record if none exists (`firstSeenAt`, `firstUserId`, `approvalState: 'auto'`, save with `announce: false`), so the client becomes a real, editable row instead of a synthetic one.
- [ ] `server/utils/clientIdMetadata.js` + `oauthClientResolver.js`: `oauth.cimd.blockedClientHosts`, checked **before** the allowlist and before any network call, same `hostMatchesPattern` semantics; layer `record.active` and host blocking into `resolveOAuthClient()` and `buildPolicyCimdClient()`.
- [ ] Blocked-client refusal: plain-HTML page at the authorize endpoint (sibling of `renderGroupDeniedPage`) naming the client and saying to contact an administrator; `401 invalid_client` on `/mcp`, identical in shape to a suspended stored client.
- [ ] `server/migrations/V110__add_oauth_cimd_governance.js` (use the `create-migration` skill; confirm the number at implementation time): `oauth.cimd.blockedClientHosts: []`, `oauth.cimd.approvalMode: 'approval'`. Seed the same in `server/defaults/config/platform.json`. The setting is **inert until Phase 3** reads it, which keeps the phases independently shippable.
- [ ] `AdminOAuthClientsPage.jsx`: CIMD rows become real rows — `Client metadata` kind badge, document URL, connection count, first seen, last used; actions **Block / Unblock** and **Revoke all connections**. Blocking revokes the client's connections automatically (a block that leaves live refresh tokens is not a block) and the confirm dialog says so, with the existing caveat that access tokens already issued expire within `tokenExpirationMinutes`. Same action on `AdminOAuthConnectionsPage.jsx` while filtered to one client.
- [ ] Audit events: `oauth.client.discovered`, `oauth.client.blocked`, `oauth.client.unblocked`, `oauth.connection.revokedBulk` (client + count). Align with `server/validators/auditEntrySchema.js`.
- [ ] Tests: bulk revoke clears consents *and* refresh tokens for every user of a client; a blocked client is refused at authorize and at `/mcp`; `blockedClientHosts` wins over `allowedClientHosts` and causes no outbound request; a discovery record is written once and is idempotent.
- [ ] `docs/mcp-integration.md` (governance table + troubleshooting), release note via `/document-feature`.

### Phase 2 — Per-client policy and request-path enforcement

- [ ] `server/utils/oauthClientPolicy.js`: the field-by-field layering from principle 4 plus `isUserAllowedByGroups`, moved out of `oauthAuthorize.js` so authorize, token refresh and `mcpAuth` share one implementation.
- [ ] Layer `allowedGroups`, `allowedApps`, `allowedModels`, `allowedPrompts`, `scopes` and `tokenExpirationMinutes` from the record over the `platform.oauth.cimd` defaults in both resolver entry points.
- [ ] `server/middleware/mcpAuth.js`: after resolving the client, run the group check against the token's groups; `403 access_denied` when the client's *current* `allowedGroups` no longer admits them. Narrowing a client's groups then kills live tokens within one access-token lifetime.
- [ ] `server/routes/oauth.js` refresh grant: same group check against `tokenData.userGroups` → `invalid_grant`; intersect the stored `scopes` with the client's current grantable scopes and drop what is no longer allowed, so narrowing `allowedScopes` narrows live connections too.
- [ ] Refresh grant, **local** users only: re-read live group membership from the user store and refresh the snapshot. OIDC/proxy snapshots stand — see *Honest limits*.
- [ ] `AdminOAuthClientEditPage.jsx`: a CIMD variant — identity fields (name, redirect URIs, grant types, secret rotation) read-only with a note that they come from the document at `<url>`; only policy fields editable; `trusted` / `consentRequired` not rendered at all.
- [ ] Tests: per-client `allowedGroups` restricts one CIMD client while another on the same host is unaffected; per-client `allowedApps` narrows only that client's `getAppsForUser` result; a client-policy change is felt on the next `/mcp` call and on the next refresh; an unset field still inherits the global default.
- [ ] Docs + release note.

### Phase 3 — The approval gate (unknown clients are blocked)

- [ ] `oauth.cimd.approvalMode`: **`approval` (default)** | `auto`. Under `approval`, the first authorization by a `client_id` with no record writes one with `approvalState: 'pending'`, `active: false`, `firstSeenAt`, `firstUserId`, and **refuses** the authorization. `auto` keeps the pre-change behaviour for installations that want the host allowlist to be the whole decision.
- [ ] Refusal page: names the client and its host and tells the user to ask an administrator to allow it — the page is the feature, so the user knows what to request and the admin gets a named row rather than a surprise in a connections list later.
- [ ] `AdminOAuthClientsPage.jsx`: pending rows sort first with a distinct badge and an **Approve** action (`approvalState: 'approved'`, `active: true`); audit `oauth.client.approved`, `oauth.client.policyUpdated`.
- [ ] **Grandfathering migration — ships in the same release as the enforcement.** Reads the consent store and writes an `approvalState: 'approved'`, `active: true` record (`approvedBy: 'migration'`, `firstSeenAt` from the earliest grant) for every distinct CIMD `client_id` that already has at least one connection. Clients nobody has connected through are deliberately *not* created — those are exactly the ones that should have to be approved. Without this, the release that turns the gate on disconnects everyone.
- [ ] `AdminMcpGatewayPage.jsx` → *Client identification*: approval-mode selector (*Require approval* — the default — / *Connect automatically*) and a blocked-hosts input beside the trusted-hosts one. i18n keys in `shared/i18n/en.json` and `de.json`.
- [ ] Tests: a never-seen client is refused and its record is pending; approving it lets the same flow through; the grandfathering migration approves exactly the clients with existing connections and creates nothing else; an upgrade disconnects nobody.
- [ ] Docs (operator workflow) + release note.

## Honest limits — to document, not paper over

| Change | Takes effect |
|---|---|
| Client blocked / deleted / CIMD disabled / host removed | Next `/mcp` request — at most one access-token lifetime |
| Client `allowedGroups`, `allowedApps`, scopes narrowed | Next `/mcp` request — at most one access-token lifetime |
| Admin revokes a connection | Next refresh; an existing access token lives out its lifetime |
| **Local** user leaves a group | Next refresh (snapshot re-read) |
| **OIDC / proxy** user leaves a group | Next interactive sign-in, or when an admin revokes the connection |

The last row is inherent: iHub does not hold the IdP's group graph and cannot poll it. The mitigations are an admin revoke and `refreshTokenExpirationDays` (default 30).

## Alternatives considered (and rejected)

- **Per-client policy in `platform.json` (`oauth.cimd.clients[]`)** — platform changes need a restart, so blocking would not be immediate, which is the one property a block must have; it also splits client administration across two admin surfaces.
- **A new `contents/config/oauth-client-policies.json`** — a second store to join on every resolve, with its own migration, loader, cache key and admin page, holding fields the first store already defines. The only argument for it is purity about "CIMD stores nothing", and that purity is about *identity*, which principle 1 preserves.
- **Materialising a full client record on first sight** (name, redirect URIs and all) — re-creates the DCR sprawl #2358 removed, and a stale `redirect_uris` copy would diverge from the document, breaking login for everyone the day the vendor adds a callback URL.
- **Blocking by revoking consents only** — lasts until the user reconnects, which for a client that re-authorizes automatically is seconds.

## Acceptance criteria

- [ ] **Revoke all connections** on a client clears every consent and every refresh token for that client in one action; the count is shown before confirming and audited after.
- [ ] Blocking a CIMD client revokes its connections and makes `/mcp` answer `401` for its existing tokens on the next request; a new authorization attempt is refused with a named page, and the user cannot reconnect until it is unblocked.
- [ ] A host in `oauth.cimd.blockedClientHosts` is refused even while it is in `allowedClientHosts`, and causes **no** outbound request.
- [ ] With the shipped default, a `client_id` that has never been approved is refused at authorize even though its host is allowed; the user sees what to ask their administrator for; the client shows as pending in Admin → OAuth → Clients, and **Approve** lets the very same flow complete.
- [ ] Upgrading an installation where users are already connected disconnects nobody: the migration approves exactly the CIMD clients that already have connections, and creates no record for any other.
- [ ] Per-client `allowedGroups` on `https://claude.ai/oauth/claude-code-client-metadata` restricts Claude Code to that group while another CIMD client on `claude.ai` is unaffected; a user outside the group is refused at authorize **and** on `/mcp` with an existing token.
- [ ] Per-client `allowedApps` / `allowedModels` / `allowedPrompts` narrow what that one client sees over MCP, intersected with the user's own permissions, without affecting other clients; a field left unset still inherits `platform.oauth.cimd`.
- [ ] Narrowing a client's grantable scopes narrows live connections at the next refresh rather than only future ones.
- [ ] Admin → OAuth → Clients shows CIMD clients as real, editable rows (kind badge, document URL, connection count, first seen, last used) with Block / Unblock / Revoke all / Approve; identity fields are read-only on the edit page and `trusted` cannot be set by any path — approving a client never makes it trusted, and consent is still required.
- [ ] Existing clients, consents and refresh tokens keep working across the upgrade; nothing is revoked or deleted without an explicit admin action.
- [ ] Audit log carries the new event types with actor, client and count.
- [ ] `npm run lint:fix && npm run format:fix` clean; new tests in `npm run test:quick`; `timeout 10s node server/server.js` boots; docs and release notes updated.

## Decisions (proposed defaults — proceed unless overruled)

1. `approvalMode` defaults to **`approval`**: an unknown client is blocked and the user is told to ask an administrator. The grandfathering migration is what keeps that safe to ship.
2. Policy records live in `oauth-clients.json`, not a new store, and never carry identity fields.
3. Blocking a client always revokes its connections — there is no "block but keep the tokens".
4. `refreshTokenExpirationDays` keeps its 30-day default here; shortening it is a separate discussion.

## References

- #2358 — One OAuth client for Claude: CIMD, DCR de-duplication, connections (PR #2360) · #1461 — MCP gateway origin
- `docs/mcp-integration.md` §Client ID Metadata Documents (CIMD) · `docs/oauth-authorization-code.md` §URL client IDs
- OAuth Client ID Metadata Document draft: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00
- MCP specification 2025-11-25, *Authorization*: https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization

manzke:
Closes #2421. Implements all three phases from the issue and ``[the concept doc](https://github.com/intrafind/ihub-apps/blob/claude/modest-bardeen-izz074/concepts/2026-09-17%20MCP%20Client%20Governance%20%E2%80%94%20Per-Client%20Policy%2C%20Blocking%20and%20Connection%20Revocation.md)`` in one change, because the resolver evaluates all of them on the same code path — splitting it would mean writing the `active` predicate three times.

## The problem

A CIMD client's `client_id` is the URL of a document it publishes, and iHub stored no record for it. That left one lever for every such client — the trusted-host list — while Claude web, Claude Desktop, Claude Code and Cowork all publish under `claude.ai` with *different* `client_id` URLs. And policy was enforced at consent time only: the refresh grant re-stamped the group snapshot frozen at first authorization, `mcpAuth` never looked at `allowedGroups`, so **removing a user from the allowed group did not end their MCP access**.

## The split that makes it work

**Identity from the document, policy from the record.** `client_name`, `redirect_uris`, `grant_types` and `token_endpoint_auth_method` still come from the document on every authorization and are never stored — CIMD's "nothing stored" property is about identity and is untouched. What is stored, in the existing `oauth-clients.json` keyed by the exact URL, is the set of administrator decisions. Those records carry `metadata.cimd: true`, never a secret, and `consentRequired: true` / `trusted: false` locked by every write path: approving a client is not trusting it.

## What changed

**Revocation and blocking**
- `revokeConnectionsForClient` clears every consent and every refresh token for one client in one action, built on the existing `revokeConnection`. The user list is the union of both stores, so a user whose consent lapsed but whose refresh token is live is caught too.
- `DELETE /api/admin/oauth/clients/:encodedClientId/connections` and `PUT /api/admin/oauth/clients/cimd/:encodedClientId`. Client ids travel base64url-encoded — they are store keys, never path segments, which is why `validateIdForPath` cannot be used.
- Blocking revokes the client's connections in the same action. A block that leaves live refresh tokens behind is not a block.
- `oauth.cimd.blockedClientHosts` is checked **before** the allowlist and before any network call, so a blocked host causes no outbound request.

**Per-client policy, enforced on the request path**
- `server/utils/oauthClientPolicy.js` holds the field-by-field layering (`record[field] ?? platform.oauth.cimd[field]`) and the group check that authorize, the refresh grant and `mcpAuth` now share — previously three different answers, one of which was "no answer".
- `mcpAuth` re-checks the client's *current* `allowedGroups` against the token's groups → `403 access_denied`. The refresh grant does the same → `invalid_grant`, and intersects the granted scopes with what the client may still be granted.
- A **local** user's groups are re-read from the user store on refresh. OIDC/proxy snapshots stand — documented, not papered over.
- `active` is computed on every request from five independently revocable conditions, which is what makes each one a kill switch rather than a gate only new connections pass.

**The approval gate**
- `oauth.cimd.approvalMode` defaults to `approval`: passing the host allowlist makes a client *eligible*, not allowed. An unapproved `client_id` is refused with a page naming the client and telling the user to ask an administrator, and becomes a pending row with an **Approve** action. `auto` keeps the previous behaviour and still writes a real, editable row.
- V112 seeds the settings; **V113 approves exactly the clients that already have a connection**, so the upgrade disconnects nobody, and creates nothing for clients nobody has connected through — those are precisely the ones that should have to be approved.

**Admin UI** — CIMD clients are real rows on Admin → OAuth → Clients (kind badge, document URL, connection count, first seen, last used) with Approve / Block / Unblock / Revoke all / Edit policy; revoke-all on the Connections page; approval-mode selector and blocked-hosts field on the MCP gateway page; a CIMD edit page whose identity fields are read-only and which never renders `trusted` or `consentRequired`. en/de.

## Worth a second look

- **Ordering of the refusals.** Everything decidable without the document (CIMD off, host blocked, host not allowed, client blocked, record already pending) is evaluated *before* the fetch, so those refusals cost no outbound request. The approval check for a client with *no* record runs after the fetch — the host is already allowlisted by then, so it costs no new exposure, and it is what lets the refusal page and the pending row name the software the user actually tried.
- **Scope intersection on refresh** narrows connections that already exist. The authorize endpoint does not intersect requested scopes against the client's registered scopes, so a connection granted scopes outside its client's list will have them dropped at the next refresh. That is the behaviour the issue asks for, but it is a change for existing stored-client connections too, not only CIMD ones.
- **`validateClientCredentials` now refuses a record with no stored secret.** Without it, posting a CIMD client's URL with any secret to `/api/oauth/token` reaches `bcrypt.compare(secret, null)`, which throws rather than returning false.
- **A pending record carries `active: false`**, so `evaluateCimdAccess` checks `approvalState === 'pending'` before the block check — otherwise a client waiting for approval would tell its user it was "blocked by an administrator".

## Migration numbering

Main took V111 for the Qwant search provider while this branch was using it, so both migrations here moved up one and kept their order — the approval gate has to be seeded before the migration that grandfathers clients through it:

- `V111__add_oauth_cimd_governance` → **V112**
- `V112__grandfather_connected_cimd_clients` → **V113**

Anyone who ran this branch before the merge has the old numbers in their history, and the stale V111 entry *shadows Qwant's*, so both renames are registered in `RENAMED_MIGRATIONS` the way the V094–V096 storage set and Qwant's own V110 → V111 move already are. Verified against a history in that exact state: both entries reconcile and Qwant's V111 then applies. Tests cover the chain, since the governance migration moves onto the number the grandfathering one is vacating — a rule matching on version alone would rewrite the same entry twice.

## Verification

- `npm run lint` / `npm run format` clean (0 errors).
- `npm run test:quick` passes on the merged tree; adds `server/tests/oauth-cimd-governance.test.js` (32 tests), `migration-v112.test.js`, `migration-v113.test.js` and two `reconcileRenamedMigrations` cases, all wired into `test:oauth` and `test:migrations`.
- `node server/server.js` boots; V112 applies, V113 skips on an installation with no consent store.
- `vite build` succeeds.
- One pre-existing test changed: `oauth-cimd.test.js` now sets `approvalMode: 'auto'` in its platform helper, because it tests identity and the host allowlist, and the new default would refuse every client in it for want of an approval record.

Docs: `docs/mcp-integration.md` gains the governance section, the "when does a change take effect" table and troubleshooting entries. Release notes in `docs/releases/next/` (feature + breaking change, the latter with the **Before upgrading** note).

https://claude.ai/code/session_01JoCuo9347QBCcMvwEQvknk


manzke:
## CodeQL: 14 × `js/missing-rate-limiting` — I believe these are false positives, and I can't clear them from inside the PR

**What's failing:** the CodeQL check, with *"14 new alerts including 14 high severity security vulnerabilities"*, all `js/missing-rate-limiting` ("This route handler performs authorization, but is not rate-limited"). `test-quick` was also failing; that one was mine and is fixed in 6b22987.

**Why I don't think it's a real defect.** Every route this PR adds sits under `/api/admin` or `/api/oauth`, and both prefixes have a rate limiter mounted *before* the routes are registered:

- `server/middleware/setup.js:628` — `app.use(buildApiPath('/admin'), rateLimiters.adminApiLimiter)`
- `server/middleware/setup.js:631` — `app.use(buildApiPath('/oauth'), rateLimiters.oauthApiLimiter)` (50 requests / 15 min / IP, `skipFailedRequests: false`)
- `server/server.js:614` calls `setupMiddleware(app, …)`; `server/server.js:651` calls `registerAdminRoutes(app)` — so the mounts are in place first.

CodeQL can't connect an `app.use` mount in one file to a route registration in another, which is a known limitation of this query. The one alert it surfaced as an inline comment is on a **test harness** — `app.get('/mcp', mcpAuth, …)` inside `server/tests/oauth-cimd-governance.test.js` — where rate limiting is meaningless.

**Caveat, stated plainly:** only 1 of the 14 was posted as an inline comment, and I have no way to enumerate the other 13 from here. I'm inferring the rest are the four new handlers in `server/routes/admin/oauthCimdClients.js` plus the other Express apps in that test file, but I haven't verified that. If any of them turns out to be on a path that *isn't* behind those mounts, that one is a genuine bug and I'll fix it — please point me at it.

**Why I'm not "fixing" it in this PR.** Three options, none of which I think is mine to take unilaterally:

1. **Per-route limiters on the new handlers.** Redundant with the mount, and it would make these four handlers the only ones in the repo that don't rely on it — the other ~255 `adminAuth` handlers all do. A reviewer would rightly ask why.
2. **A CodeQL path exclusion for `server/tests/**`.** There is no CodeQL workflow or `.github/codeql/` config in the repo, so this is running on GitHub's **default setup** — the exclusion would be a repo *settings* change, not a file I can add here.
3. **Inline `// codeql[js/missing-rate-limiting]` suppressions.** Possible, but there's no precedent for suppression comments anywhere in the tree, and scattering them through a feature PR seems like the wrong place to establish one.

**What I need:** a call on which of those you want. Happy to implement any of them — (3) is a two-line change per site if you'd prefer the check green over the noise. Alternatively these can be dismissed in the Security tab as "used in tests" / "false positive", which is where I'd lean given the alert numbering (the linked alert is `#653`) suggests a substantial pre-existing backlog of the same rule against the handlers this PR sits beside.

I'll keep watching the PR either way.

---

PR #1911, opened 2026-07-09
Fix createSourceManager silently ignoring per-call config

manzke:
## Summary
- `createSourceManager()` caches a single `SourceManager` in module scope and only honors the `config` argument on the very first call — every later call gets the already-constructed instance, silently discarding whatever `config` it was passed.
- `PromptService.js` and `PromptNodeExecutor.js` both called it with a `filesystem.basePath` override that only ever took effect if one of them happened to run first; this "worked" only because the override is provably equal to `FileSystemHandler`'s own default. Dropped the now-redundant override at both call sites and call `createSourceManager()` with no args.
- Added a documenting comment on `createSourceManager` in `server/sources/index.js` explaining the singleton's first-call-wins behavior, so it isn't mistaken for a per-call config API in the future.
- Added a regression test (`server/tests/source-manager-singleton.test.js`) asserting that a second call with a different `filesystem.basePath` does not affect the already-constructed instance.

## Test plan
- [x] `npm run lint:fix` / `npm run format:fix` on changed files — no new warnings/errors introduced
- [x] `server/tests/source-manager-singleton.test.js` (new) passes
- [x] `server/tests/SourceResolutionService.test.js` and `server/tests/admin-sources-user-context.test.js` still pass

Closes #1794

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MKw7MBebmGMsF6SpBh6aLB

---
_Generated by [Claude Code](https://claude.ai/code/session_01MKw7MBebmGMsF6SpBh6aLB)_

PR #2474, opened 2026-09-22
One tagged message shape for every client, rendered by the server (fixes #2454)

### LINKED ISSUE #2454: Default apps (Translator, Summarizer, …) ignore the Outlook context blocks and act on the typed instruction instead
Reported in customer/field feedback on the Outlook add-in: the EN-Translator was run on a mail opened as Forward and **translated only the typed text ("hello, how are you"), not the mail body**.

## Root cause

The add-in wraps the mail in named blocks and puts the user's own text in `<user_instruction>` (`docs/outlook-add-in.md`, "What the model receives"; built in `client/src/features/office/utilities/buildChatApiMessages.js:277-280`). The whole assembly is then substituted into the app's `{{content}}`.

`server/defaults/apps/translator.json` was written for the web app and knows nothing about those blocks:

```json
"prompt": {
  "en": "Selected Language: \"{{language}}\" - Text to translate: \"{{content}}\"",
  "de": "Ausgewählte Sprache in die Übersetzt werden soll: \"{{language}}\" - Text der übersetzt werden soll: \"{{content}}\""
}
```

The model therefore receives *"Text to translate: `<current_email>…</current_email><user_instruction>hello, how are you</user_instruction>`"* and has no way of knowing which part is the subject of the task. Picking the user's sentence over the quoted mail is a reasonable reading of that prompt — the prompt is the bug.

Our own documentation says as much:

> Write app prompts against these tags. The shipped **Outlook – Reply Directly** app (`outlook-reply`) is the reference: its prompt template names the blocks, tells the model that `<user_instruction>` decides the content of the reply …

`outlook-reply` is the **only** default app that does this. Every other app a user can pick in the task pane — Translator, Summarizer, Email Composer, File Analysis, Chat — has a context-free web-app prompt.

## Suggested change

* Teach the default apps that are realistically used from Outlook to name the blocks: when `<current_email>` / `<current_meeting>` is present, that is the material to act on and `<user_instruction>` modifies *how*. Translator and Summarizer are the urgent two.
* Alternatively (or additionally), have the server inject a short standard instruction next to `<context_rules>` covering the "the blocks are the subject of the task unless the user says otherwise" case, so third-party and customer-authored apps get sane behaviour without every prompt being rewritten. `<context_rules>` already exists as the place for cross-app boundary rules.
* Document the expectation for customer-authored apps in `docs/outlook-add-in.md` — right now the guidance exists but the shipped apps do not follow it.

While in `translator.json`: it has no `upload` block, so the Translator is the one obvious document app in the pane that cannot take a document. Field feedback asked for exactly that. Adding an `upload` section (as `email-composer.json` has) is a config-only change.

manzke:
Fixes #2454.

## Why

`{{content}}` meant different things depending on the client. In Outlook and the browser extension it held client-built tag blocks; in the web app it held only the typed text, and the server put uploads *above* the app template as unescaped `[File: …]` sections. Apps written for one shape broke in the other:

- Outlook / extension → Translator: `Text to translate: "<current_email>…<user_instruction>hello, how are you</user_instruction>"`, so the model translated the note.
- Web app → Translator with an upload (the change the issue proposed): `[File: contract.docx]…` above `Text to translate: "into German please"`. Same bug.
- An email attachment could carry a raw `<user_instruction>` ahead of the real one. Only email, page and meeting fields were escaped.

## What changes

**One renderer, on the server** (`shared/promptContext.js`, called from `PromptService.processMessageTemplates`). Clients send `content` (the typed text), `hostContext` (structured open email / page / meeting and added emails, as display strings) and `fileData`. Each piece of material becomes one block, and the result goes into `{{content}}`:

```
<content type="email" origin="added">…</content>
<content type="email" origin="open">…</content>                     or type="meeting" / "page"
<content type="document" origin="attachment" name="…" format="…">…</content>
<content type="document" origin="upload" name="…" format="…">…</content>
<context_rules>…</context_rules>
<user_instruction>…</user_instruction>
```

- **One tag.** `type` says what the block is, `origin` where it came from. These are facts, not roles. Whether an added email is background (a reply) or the thing to work on ("summarize these") is up to the app's task and the user's instruction. A new host adds a `type` value, not a tag every app prompt has to learn.
- **`<context_rules>`** explains the origins and says the app's task applies to all of the material unless `<user_instruction>` narrows it. So the Translator translates an attached PDF along with the email, and customer apps with a "text to translate: {{content}}" style prompt behave sensibly without a rewrite.
- **A message that is only typed text is sent unchanged**, because in a Translator or Summarizer the typed text is the material itself.
- **Escaping.** Our tag names inside source text (email, page, meeting fields, attachment and upload text) are escaped. PDFs without a text layer are listed with `pages_as_images`. `RequestBuilder` now only attaches the page images, and merges them with the user's own images instead of replacing them.
- **Office adapter / extension.** `buildHostContext` replaces the client-side tag builder, and attachments carry `origin: 'attachment'`. `useAppChat` forwards `hostContext`. The Outlook pane's token estimate and `shared/contextUsage.js` count exactly what the server renders.
- **`@workflow` mentions** render the host context into the workflow input, since `content` no longer carries it.
- **Stored chat history** keeps the typed text, not the email body. Uploads already worked this way; the current item is re-read on every send.
- **`FileSystemHandler`** resolves an absolute `CONTENTS_DIR` instead of nesting it under the root. #1911 had made that path load-bearing (see the comments below).

**Apps + migration V122.** Translator, Summarizer, Outlook – Reply Directly and both meeting apps name the `<content>` blocks. The reply app's own `<content>` section is renamed `<reply_content>`. Translator and Summarizer stop quoting `{{content}}` and get a document `upload` section. V122 replaces a prompt with the new default only while it is exactly the text shipped before, checked by hash. The test fixtures hold those texts. An admin-edited language is left alone, and `upload` is added only where none exists.

**No backward compatibility, as agreed.** The old tags (`<current_email>`, `<pinned_emails>`, `<current_meeting>`, `<current_page>`) never shipped in a release, so this is the moment to replace them. An older, separately installed browser extension build keeps sending tags it builds itself, so it needs the update that ships with this release.

**Docs:** `docs/apps.md` has a new section, "What `{{content}}` contains", with an attribute table. `docs/outlook-add-in.md` is updated. The release notes extend the unreleased tagged-blocks breaking-change entry and add a fix entry for the Translator/Summarizer.

Not in scope: workflow agent nodes still inject their `inputFiles` as `[File: …]` (`PromptNodeExecutor`). That is a separate surface whose prompts workflow authors write themselves.

## Testing

- `server/tests/prompt-context-blocks.test.js` runs `processMessageTemplates` with the real Translator template: Outlook email + attachment, web upload, bare text, history turns, placeholder/`$` literals and forged tags.
- `server/tests/migration-v122.test.js` checks that the shipped texts become the new defaults, a fresh install is untouched, admin edits are kept, and the migration is idempotent.
- The Office unit tests are rewritten against `buildHostContext` + the shared renderer.
- `npm run test:quick` passes. Lint and format are clean.
- A server boot applies V122 to the five previously shipped app files and produces exactly the new defaults.

https://claude.ai/code/session_01EJ79z9JsD55CkhDQuNntPL


manzke:
**`test-quick` is red because of `main`, not this PR.**

The failing test is `server/tests/promptService-sources-dollar.test.js`:

> Source dollar: Access denied: path sources/dollar.md is outside allowed directory

CI runs the PR merged with the current `main`, which now includes #1911 (`70db86c`). I reproduced it with `TMPDIR` inside the checkout, as on the runner:

| Tree | Result |
| --- | --- |
| `70db86c` (`main` with #1911) | fails |
| `c5f66f4` (the commit before it) | passes |
| this PR's own head | passes |

**Cause.** #1911 dropped the `filesystem.basePath` override in `PromptService` and `PromptNodeExecutor`, assuming it equalled `FileSystemHandler`'s default. It doesn't when `CONTENTS_DIR` is an absolute path, which is what this test sets:

- the override used `path.resolve(getRootDir(), CONTENTS_DIR)`;
- the default uses `path.join(rootDir, contentsDir)`, which nests an absolute directory under the repo root. That gives `…/ihub-apps/tmp/ihub-sources-dollar-…`, the basePath in the CI log.

This affects real installs with an absolute `CONTENTS_DIR` too, not only the test.

**Proposed fix (no fix PR exists yet).** A one-line change in `server/sources/FileSystemHandler.js`. With it, the test passes on `70db86c`:

```diff
-    this.basePath = handlerConfig.basePath || path.join(rootDir, contentsDir);
+    this.basePath = handlerConfig.basePath || path.resolve(rootDir, contentsDir);
```

Using #2473's `getContentsPath()` there would work as well. I'm not re-running the job: the failure is deterministic, and a re-run would test the same merge. Once that fix is on `main`, this PR goes green.

---

manzke:
The `CONTENTS_DIR` fix from my comment above is now in this PR, as asked (a25caa3, on top of a merge of the latest `main`):

- `FileSystemHandler` uses `path.resolve(rootDir, contentsDir)`, so an absolute `CONTENTS_DIR` is kept as it is.
- With `TMPDIR` inside the checkout, as on CI, `promptService-sources-dollar` fails without the change and passes with it.
- The full `npm run test:quick` passes on the merged tree.

---

PR #1936, opened 2026-07-09
fix(auth): make AuthContext the single owner of /auth/status

manzke:
## Summary

Closes #1783.

`AuthContext` fetched `/auth/status` via two raw, uncached `apiClient.get()` calls (on mount and in the token-expiry handler), bypassing the cached/deduped `fetchAuthStatus()` helper that `PlatformConfigContext` already used for the same endpoint. That meant a fresh page load fired the endpoint from both providers independently, and it fired again on every OIDC callback and token-expiry re-check. Meanwhile `platformConfig.authenticated`/`platformConfig.user` were dead duplicate state — no component read them — and `UserAuthMenu`/`LoginForm` read auth-method flags off `platformConfig` instead of the richer `authConfig` already available from `useAuth()`.

- `AuthContext.loadAuthStatus()` / the token-expiry handler now call `fetchAuthStatus()` from the shared cache module instead of a raw `apiClient.get('/auth/status')`.
- `handleAuthGateSuccess` now clears the API cache before refetching. The auth-gate overlay (`auth-gate.js`) logs in via its own `fetch()` calls outside our cache module, so without this the post-login refetch would serve the stale pre-login cache entry — verified this regression locally (login would leave the user stuck on `/login`) and fixed it before pushing.
- Removed the unused `authenticated`/`user` fields from `PlatformConfigContext`'s `combinedConfig`.
- `UserAuthMenu.jsx` / `LoginForm.jsx` now read `authMode`, `anonymousAuth`, and `local.showDemoAccounts` from `authConfig` (via `useAuth()`) instead of `platformConfig`.

## Verification

Ran the dev server and drove it with a headless browser:
- Counted `/api/auth/status` network requests on a fresh boot: 4 → 2 (remaining 2 are React StrictMode's dev-only double-invoke; production fires once).
- Logged in as the local admin (`admin` / `password123`) end-to-end — confirmed the user menu renders "Demo Administrator", shows groups, the Admin Panel link, and Sign Out.
- Confirmed the login page's "Demo accounts" hint (driven by `authConfig.authMethods.local.showDemoAccounts`) still renders correctly.
- `npx eslint` on the four changed files: 0 errors (only pre-existing unrelated warnings).
- Client production build succeeds.

## Test plan

- [x] Fresh page load fires `/auth/status` fewer times than before (verified via network capture)
- [x] Local login via the auth-gate overlay reaches the authenticated app state (previously regressed, now fixed)
[...]

PR #2475, opened 2026-09-22
fix(auth): make AuthContext the single owner of /auth/status (supersedes #1936)

### LINKED ISSUE #1783: PlatformConfigContext duplicates auth and UI state owned by AuthContext/UIConfigContext
**Category:** architecture

### Problem
Three contexts fetch and hold overlapping server state. PlatformConfigProvider fetches /auth/status + /configs/ui + /configs/platform and stores auth fields (authenticated, user, authMethods) and UI fields inside `platformConfig`; AuthContext independently fetches '/auth/status' via a direct apiClient.get (bypassing the cached/deduped fetchAuthStatus endpoint — a second network request on boot); UIConfigProvider fetches /configs/ui again. The result is two sources of truth for the current user and for auth methods, and consumers like UserAuthMenu already hedge between them.

### Evidence
PlatformConfigContext.jsx:23-27: `const [authStatus, uiConfig, platformCfg] = await Promise.all([fetchAuthStatus(...), fetchUIConfig(...), fetchPlatformConfig(...)])` then lines 65-67 store `authenticated: authStatus.authenticated, user: authStatus.user`. AuthContext.jsx:103: `await apiClient.get('/auth/status', ...)` (uncached, separate request). UserAuthMenu.jsx:47-48: `authConfig?.anonymousAuth?.enabled ?? platformConfig?.anonymousAuth?.enabled !== false`.

### Recommended fix
Make AuthContext the single owner of /auth/status (using the cached fetchAuthStatus endpoint) and have PlatformConfigContext expose only platform features; drop user/authenticated/authMethods from platformConfig and remove the redundant fetchUIConfig in one of the two providers.

### Affected files
- `client/src/shared/contexts/PlatformConfigContext.jsx:23`
- `client/src/shared/contexts/AuthContext.jsx:103`
- `client/src/features/apps/components/AppProviders.jsx:35`

---

manzke:
## Summary

Closes #1783. Supersedes #1936, whose branch was about 2.5 months behind `main` and no longer merged. The same change is ported onto current `main`, with one refinement and a regression test.

`AuthContext` fetched `/auth/status` with raw `apiClient.get()` calls on mount and in the token-expiry handler. That skipped the cached, deduplicated `fetchAuthStatus()` helper that `PlatformConfigContext` already uses, so every page load, OIDC callback and token-expiry check hit the endpoint twice. `platformConfig.authenticated` and `platformConfig.user` were also duplicate state that no component read.

- `loadAuthStatus()` now calls `fetchAuthStatus()`. The login paths (local/LDAP credentials, token/OIDC callback) already run `clearApiCache()` before calling it, so a refresh after login still reaches the server.
- **Token expiry:** the handler calls `invalidateAuthStatusCache()` and then `fetchAuthStatus()`. #1936 used `skipCache: true` here, which left the stale "authenticated" entry in the cache for other consumers. This follows the invalidate-then-refetch pattern `PlatformConfigContext.loadConfig` adopted on `main` since then.
- **Auth-gate success:** clears the API cache before refetching. The auth-gate overlay logs in with its own `fetch()` calls outside the cache module, so without this the refetch would return the pre-login entry.
- Removed the unused `authenticated`/`user` fields from `PlatformConfigContext`'s combined config.
- `UserAuthMenu` and `LoginForm` now read `authMode`, `anonymousAuth` and `local.showDemoAccounts` from `useAuth().authConfig` instead of `platformConfig`. `authConfig`'s initial state already defines `authMode` and `anonymousAuth.enabled`, so the old `platformConfig` fallbacks never took effect. Behavior doesn't change.

**`/auth/user` call sites (raised in the triage comment):** left as they are on purpose. `loginWithToken` checks a specific token it was just handed, and `refreshUser` is an explicit refresh. Neither should be cached, and `/auth/user` has no shared cached helper or second consumer to dedupe with.

No release-notes entry: the change removes duplicate requests and has no visible effect for admins or users.

## Verification

- New `tests/unit/client/auth-status-single-owner.test.jsx` checks three things: boot goes through `fetchAuthStatus()` instead of `apiClient.get('/auth/status')`, auth-gate success clears the cache before refetching, and token expiry invalidates the cache before re-checking. All 3 fail on the old `AuthContext` and pass with this change.
- `jest tests/unit/client`: 90 suites, 837 tests, all pass.
- ESLint on the changed files: 0 errors, only warnings that were already there. Prettier is clean. The client production build succeeds.
- Dev server driven with headless Chromium: a fresh boot fires `/api/auth/status` twice, both from React StrictMode's dev-only double mount. The login page shows the demo-accounts hint, read from the new `authConfig` source. Local login as `admin` / `password123` authenticates and lands on the admin-only setup wizard, and there were no page errors.

https://claude.ai/code/session_0196GcdWEtMR5Bb6zuMgzrup

---
_Generated by [Claude Code](https://claude.ai/code/session_0196GcdWEtMR5Bb6zuMgzrup)_


manzke:
**`test-quick` is red, but the failure isn't from this PR.** It is `server/tests/promptService-sources-dollar.test.js` under `test:prompts`, and this PR only changes client files.

It comes from the base branch: #1911 (`70db86c`, merged to `main` minutes before this PR opened) removed PromptService's `createSourceManager({ filesystem: { basePath: path.resolve(getRootDir(), config.CONTENTS_DIR) } })` override. That leaves `FileSystemHandler`'s default, `path.join(rootDir, contentsDir)`. The two only match for a relative `CONTENTS_DIR`. With an absolute one, which is what that test uses (an `os.tmpdir()` path), `path.join` turns it into `<repo>/tmp/ihub-sources-…`. The source then fails with "outside allowed directory" and `{{sources}}` renders empty. The failure is deterministic, not a flake: it reproduces locally on `main`, and restoring the pre-#1911 `PromptService.js` makes it pass.

I found no open PR that fixes this. Proposed patch, tested locally: `npm run test:prompts` exits 1 without it and 0 with it.

```diff
--- a/server/sources/FileSystemHandler.js
+++ b/server/sources/FileSystemHandler.js
-    this.basePath = handlerConfig.basePath || path.join(rootDir, contentsDir);
+    this.basePath = handlerConfig.basePath || path.resolve(rootDir, contentsDir);
```

This makes the handler default match what the removed override did, so the "equal to the default" premise of #1911 holds again. I'm not adding it here so this PR doesn't grow beyond its scope. It can go in a small follow-up on `main`, and this PR's CI will go green once that lands.

---

PR #2476, opened 2026-09-22
refactor(config): load every configCache key through one loader registry (supersedes #1934)

### LINKED ISSUE #1785: configCache: triplicated per-type load logic, inconsistent return shapes, and dead/broken getWithFallback

manzke:
## Summary

Fixes #1785. Supersedes #1934, rebuilt on current `main`: that branch had drifted about 2.5 months against a file that kept growing, so it could not merge cleanly.

`server/configCache.js` had a separate per-type copy of the loading logic in `initialize()`, in `_reloadEntry()`, and again in each `refreshXCache()` wrapper. The copies had drifted in ways users could see. This PR puts all of it in one declarative `CONFIG_LOADERS` table that both boot and refresh read.

### Real bugs this fixes (each has a regression test that fails on `main`)

- **Voice input broke after a platform save.** The refresh path read `platform.json` without decrypting `speech.realtime.apiKey` / `speech.azure.subscriptionKey`; only boot decrypted them. After any admin save, or the first TTL tick once the keys are encrypted, the cache held the `ENC[...]` string. vLLM Realtime and the Azure token broker then sent that string upstream until the server restarted.
- **Edits made on disk stopped being picked up.** The apps/models/prompts/workflows/agents branches of `_reloadEntry()` never re-armed the TTL timer when nothing had changed. The first quiet tick ended the refresh chain, which is the same bug `config-cache-refresh-chain.test.js` pins for `groups.json`.
- **Getters returned inconsistent shapes.** `getApps()` / `getTools()` / `getPrompts()` returned a bare `[]` on a cold cache but `{ data, etag }` otherwise. They now always return `{ data, etag }`.

### Other changes

- The registry also covers `sources`, `providers`, `registries` and `installations`, as asked in the triage comment. Keys not listed fall back to `loadJson(key)` under the label "config", so there is a single load path.
- A reload now compares the ETag of the **resolved** data (after env-var and `IHUB_*` overrides) with the cached ETag. It stores only when something changed, and always re-arms the timer, including after a failed reload. Before, raw and resolved ETags were compared, so any entry with an env placeholder looked changed on every tick.
- The `refreshXCache()` wrappers delegate to `refreshCacheEntry()`, which still announces the change to the cluster.
- `sourceDependencyTracker` iterated the `{ data, etag }` object from `getApps()` as if it were an array. It now destructures `.data`. (The module has no importers today, so this had no visible effect.)
- Removed the unreachable `getWithFallback()`, which had no callers.
- Dropped the redundant `|| {}` around `getTools()` in `requestThrottler.js`.
- Added two `docs/releases/next/fixes.md` entries, one for the voice keys and one for on-disk edits.

A correction to #1934's description: overwriting an entry with identical data never actually changed its ETag, because the ETag is a hash of the content. The consolidation's real value is the drift bugs listed above.

(An earlier commit here ported a `FileSystemHandler` `path.resolve` fix for `main`'s red `test-quick`. `main` has since landed the same fix, and the merge took `main`'s version, so this PR no longer changes that file.)

## Verification

- New `server/tests/config-cache-loaders.test.js` (added to `test:config-store`): 4/4 pass here and all 4 fail on unmodified `main`.
- Full `npm run test:quick` passes locally on the merged head (exit 0; 1315 Jest unit tests plus the node test suites).
- `node server/server.js` boots with every config type logging "Cached …" with the expected counts, and 0 error-level log lines.
- eslint and prettier are clean on all changed files.

https://claude.ai/code/session_019dWgh3oReaz7fHEQ6GXg3c


manzke:
`test-quick` failed in `server/tests/promptService-sources-dollar.test.js` ("Access denied: path sources/dollar.md is outside allowed directory"). The failure isn't from this PR: it reproduces on `main` at 70db86c.

Cause: #1911 removed PromptService's `path.resolve(getRootDir(), CONTENTS_DIR)` base-path override and relied on `FileSystemHandler`'s default, which used `path.join` instead. The two differ when `CONTENTS_DIR` is absolute, as it is in that test.

No fix existed yet, so I ported a one-line fix into this PR in adab15e: `FileSystemHandler` now uses `path.resolve` for its default. It becomes a no-op once `main` carries an equivalent fix. The full `npm run test:quick` passes locally.

---

PR #1937, opened 2026-07-09
refactor(client): stop shared/ modules from importing features/ (#1780)

manzke:
## Summary

Closes #1780.

Two `client/src/shared/` modules imported from `features/`, inverting the project's documented layering (features depend on shared, never the reverse):

- **`useEventSource.js`** (the app-wide chat SSE hook, used by every chat stream — not just Office) unconditionally imported `getRefreshToken`/`refreshTokenOrExpireSession` from `features/office/api/officeAuth` and hardcoded the Office add-in's `office_ihubtoken` localStorage key into its auth-header logic. A change to Office's token lifecycle could silently affect the plain SPA chat path.
- **`DynamicLanguageEditor.jsx`** (used in 17+ places, including non-admin contexts) unconditionally imported `useFormValidationErrors` from `features/admin/components/formValidationContext`.

### Changes

- `useEventSource` now accepts optional `getAuthHeaders`/`onUnauthorized` options instead of baking in Office-specific logic. Default behavior (no options) is unchanged for the plain SPA path: read `authToken` from `localStorage`, no 401 retry.
- `useAppChat` accepts and forwards `getAuthHeaders`/`onUnauthorized` straight through to `useEventSource`.
- `useOfficeChatAdapter` now supplies its own `getAuthHeaders` (Office token → fallback to `authToken`) and `onUnauthorized` (silent refresh via `officeAuth`), preserving its existing behavior but sourced from the Office feature instead of baked into `shared/`.
- `features/admin/components/formValidationContext.jsx` moved to `client/src/shared/contexts/formValidationContext.jsx` (pure relocation, no logic change), with its 6 importers (`DynamicLanguageEditor` + 5 admin form editors) updated to the new path.

This is a pure refactor — no behavior change for either the plain SPA or the Office add-in chat paths.

## Test plan

- [x] `npm run lint:fix` — no new warnings/errors introduced (pre-existing unrelated warnings only)
- [x] `npm run format:fix`
- [x] `client` production build succeeds (`vite build`)
- [ ] Manual verification in a running dev server: plain chat send/stream with only `authToken` set; Office add-in chat send/stream and silent refresh-on-401 retry; admin form editors (App/Model/Prompt/User/Group) still show per-language validation errors

🤖 Generated with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_01PsXJ3us2fRzKZzguXEngV2

---
_Generated by [Claude Code](https://claude.ai/code/session_01PsXJ3us2fRzKZzguXEngV2)_

Issue #1780, opened 2026-07-06
shared/ modules import from features/ — layering inversion bakes office/admin specifics into shared code


manzke:
**Severity:** medium · **Confidence:** suggestion · **Effort:** medium
**Category:** architecture

### Problem
The documented architecture is features depend on shared, not the reverse. Two shared modules violate this: useEventSource.js (the app-wide chat SSE hook) imports getRefreshToken/refreshTokenOrExpireSession from features/office/api/officeAuth and hardcodes the Office add-in token key 'office_ihubtoken' into its auth-header logic, so the core chat streaming path silently depends on the Office feature's token lifecycle. DynamicLanguageEditor.jsx (used in 17 places) imports useFormValidationErrors from features/admin/components/formValidationContext. This makes features impossible to remove/lazy-load independently.

### Evidence
useEventSource.js:4: `import { getRefreshToken, refreshTokenOrExpireSession } from '../../features/office/api/officeAuth';` and :101-103: `const token = localStorage.getItem('office_ihubtoken') || localStorage.getItem('authToken')`. DynamicLanguageEditor.jsx:5: `import { useFormValidationErrors } from '../../features/admin/components/formValidationContext';`.

### Recommended fix
Invert the dependencies: let useEventSource accept an optional getAuthHeaders/onUnauthorized callback that the Office adapter supplies, and move formValidationContext into shared/ (or pass validationErrors as a prop, which the component already supports).

### Affected files
- `client/src/shared/hooks/useEventSource.js:4`
- `client/src/shared/hooks/useEventSource.js:102`
- `client/src/shared/components/DynamicLanguageEditor.jsx:5`

---
<sub>From the 2026-07-06 comprehensive review · tracking #1677 · report `concepts/2026-07-06 Comprehensive Code Review.md`</sub>

manzke:
## Implementation Plan

**What:** Two `client/src/shared/` modules currently import from `features/` (a layering inversion documented in `docs/` architecture as forbidden), so core chat streaming and a widely-reused form component both carry feature-specific dependencies they shouldn't need.

[...]

PR #2477, opened 2026-09-22
refactor(client): stop shared/ modules from importing features/ (#1780, supersedes #1937)

manzke:
## Summary

Refs #1780. Replaces #1937, which conflicts with `main` because the SSE transport has since moved out of `useEventSource.js` into `client/src/shared/utils/openSseStream.js`. That new module now has the same layering problem. This PR applies the #1937 fix to where the code lives today.

### SSE auth (`shared/utils/openSseStream.js`)

`openSseStream` imported `getRefreshToken` and `refreshTokenOrExpireSession` from `features/office/api/officeAuth` and hardcoded the `office_ihubtoken` key. #1937 passed `getAuthHeaders`/`onUnauthorized` down through `useAppChat` → `useEventSource`. That no longer fits: `openSseStream` / `fetchWithAuthRetry` serve chat, the chat re-sync, `useRunStream` (workflows and agent runs) and `OcrPage`. Passing options through all of them would still leave some streams without Office auth.

This PR adds a registration hook in the shared module instead:

- `setSseAuthProvider({ getToken, onUnauthorized })`. `getSseAuthHeaders()` prefers the provider's token and falls back to `authToken`. `fetchWithAuthRetry()` retries a 401 once, and only when the provider's `onUnauthorized()` resolves `true`.
- `installOfficeAuthInterceptor` registers the Office provider (Office access token, plus a silent refresh keyed off `getRefreshToken()`) next to its existing Axios interceptors. The Office taskpane, the browser extension and the Nextcloud embed all already call it, so every SSE stream in those hosts keeps the Bearer token and the 401 refresh-and-retry.
- In the plain SPA, SSE now sends only `authToken` and doesn't attempt an Office refresh. That matches the SPA's Axios clients, which never used the Office token either.

### `formValidationContext`

Moved from `features/admin/components/` to `shared/contexts/`. This is a pure relocation. `DynamicLanguageEditor` and the five admin form editors now import it from the new path.

### Ported CI fix (not part of this refactor)

`server/sources/FileSystemHandler.js`: `path.join` → `path.resolve` for the default basePath. Since #1911, `main` fails `test-quick` (`promptService-sources-dollar.test.js`) because an absolute `CONTENTS_DIR` gets nested under the installation root. This is the same one-line change as in #2474, and it becomes a no-op once that PR lands.

### Not in scope

The triage comment on #1937 lists two newer violations of the same rule: `Layout.jsx` imports `UserAuthMenu` and `useEmbeddedHostKind`, and `AppSidebar.jsx` imports `ChatTitleEditor` and `UserAuthMenu`. Fixing them means relocating or slotting whole feature components, so they're left for a follow-up.

## Test plan

- [x] New `tests/unit/client/openSseStream-auth-provider.test.jsx` (9 tests) covers: no provider → only `authToken` and no retry; provider token takes precedence and falls back to `authToken`; retry with the refreshed token; no retry when the provider declines; a failed refresh propagates; non-401 responses skip the provider.
- [x] `npm run test:ui`: 90 suites, 843 tests pass
- [x] `npm run test:quick` passes
- [x] `npm run lint:fix` / `npm run format:fix`: no new errors
- [x] `vite build` succeeds
- [ ] Manual check: Office add-in chat stream plus silent refresh on 401; admin form editors still show per-language validation errors

https://claude.ai/code/session_01X1RizgXNE5hrDV1ertXt8j


manzke:
`test-quick` failed on `server/tests/promptService-sources-dollar.test.js`. The failure comes from `main`, not from this PR:

- #1911 (70db86c) removed PromptService's `path.resolve(getRootDir(), config.CONTENTS_DIR)` basePath override.
- The fallback, `FileSystemHandler`'s default, uses `path.join`, which puts an absolute `CONTENTS_DIR` under the installation root. The test runs with an absolute `CONTENTS_DIR`, so its fixture source gets rejected as "outside allowed directory".

The failure reproduces locally on this branch without the fix. #2474 already carries a one-line fix (`path.join` → `path.resolve`), and I've ported it here in 3c1b3625. It becomes a no-op once #2474 lands. `npm run test:quick` now passes locally.

---

